PR #4, opened 2023-09-23
Moar SPEEEED

masmullin2000:
Rayon go brrrrr


jinyus (changes requested):
Hey, thanks for the pr. I did get a speed up to 20ms but I'd like to keep the single core solution separate so could you put the changes in a rust_parallel directory?

PR #6, opened 2023-09-23
go perf fix

nandesu-utils:
i noticed that in go in comparison to rust a mapping was used for storing post count instead of an array which likely led to worse performance

i replaced a mapping with an array it (and also put `log.Panicln` instead of fatal log and few more fixes), i've got like 9x-ish perf boost up to 0.06s (previously was somewhat around 0.56s)

### DIFF
--- a/go/main.go
+++ b/go/main.go
@@ -26,46 +26,47 @@ type PostWithSharedTags struct {
 }
 
 func main() {
-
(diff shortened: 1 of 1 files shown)


jinyus:
Thanks, this optimization was proposed by the rust community and I didn't get a chance to port it to go

PR #7, opened 2023-09-23
Improved performance a little with generics

DrBlury:


Mxrk:
Using generics should indeed be faster:
https://github.com/ugurcsen/gods-generic#testing-and-benchmarking

jinyus:
PS: the call to Clear is not needed because I limit the size of the heap and it's is being drained on every iteration.

Mxrk:
https://github.com/jinyus/related_post_gen/blob/5b242fc36629e32504a9a97b14641a84c374f1f7/go/main.go#L55C2-L55C62

On reddit you said this is making a difference for you, is this still the case?

PR #10, opened 2023-09-24
Use indexes, heap-based top k for python as the other languages do

dave-andersen:
8.1 -> 1.35 seconds, approximately, on M1.


jinyus:
Thanks, the python was neglected :)

PR #11, opened 2023-09-24
Used numpy for much faster python implementation

Copper280z:
This numpy implementation runs in ~0.72sec on my m2 air.

Edit: Knocked another 100ms off with array broadcasting, main now runs in 0.65 sec, 0.54sec without IO.


jinyus:
Thanks for the PR. I got roughly the same speed on my machine

PR #15, opened 2023-09-25
Even slower python and jq implementations

fizmat:
- pandas - dataframes, even slower than python (~5s) 
- polars - newer pandas-like dataframes, almost not slower (~1.4s)
- jq - stream editor for json, result unexpectedly slow, maybe someone can improve it (~50s)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ related_*.json
 .DS_Store
 .vscode/settings.json
 venv/
+.idea/
+*.py.pclprof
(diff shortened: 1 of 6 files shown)


jinyus:
Hey, thanks for the PR. I'm not adding slower implementations for languages to keep to repo lean. I would add the jq so let me know if you're going to modify the PR or I can add it manually.

fizmat:
Hey, thanks for looking it over. I'm going to experiment with improvements to all the various python solutions I want to try (in my fork for now).

And jq I will just PR separately. I'm sure there's a way to make it faster. I know just enough to make it work at all. But then again, I don't think it was designed for such not-really-streamable processing.

PR #19, opened 2023-09-26
Julia implementations

atthom:
Two implementation in Julia based on the Python code.
I tried to be as close as possible as the python code. 
Om my CPU, naive implementation takes 1.2s and the one based on numpy takes 450ms


jinyus:
Hey, thanks for the PR, could you modify the code to just execute the script and not run the benchmark? I know nothing about julia so it'd be great if you edited the run.sh file to include a function for it.

atthom:
Sure, I will update the run.sh

jinyus:
Hey, also remember to install the deps in the script. Similar to how the run_python_np function is done.

jinyus:
is there a way to install the needed packages from the command line? I had to open the julia shell in stall them

atthom:
I updated the script and the run.sh to install the packages

PR #24, opened 2023-09-27
added new (faster) duckdb (1.14s) and native python version

azrethos:
```
➜  python -mtimeit -n1 -r1 -t -s "from duck import main" "main()"

1 loop, best of 1: 1.14 sec per loop
```


jinyus:
Hey, thanks for the PR. I'm willing to merge the duckdb version but it only outputs the id of the related posts which breaks compatibility with the other versions. 

The python version is slower than the current one `3.7s vs 1.8s` so I will keep the current impl.

python/related_duckdb.py
```diff
@@ -19,18 +19,20 @@
...
+                    SELECT
+                        JSON_OBJECT('_id', rp._id, 'title', rp.title, 'tags', rp.tags)
+                    FROM related_counts rc
+                    JOIN posts rp ON rp._id = rc.related_post
+                    WHERE rc.source_post = p._id
+                    ORDER BY rc.relation_count DESC
+                    LIMIT 5
+                ) AS related
+            FROM posts p
+        ) TO '../related_posts_duckdb.jsonl';""",
```
azrethos, line 35: technically this produces JSONL https://jsonlines.org/

azrethos: `python -mtimeit -n1 -r1 -t -s "from duck import main" "main()"`


1 loop, best of 1: 1.98 sec per loop

azrethos:
I went ahead and updated it.

Note, it produces jsonl which appears to be the superior format for this type of data 

https://jsonlines.org/

jinyus:
Looks good

PR #37, opened 2023-09-30
Cutting down the number of threads in Rayon

masmullin2000:
Rayon seems to perform significantly better by only using the amount of physical cpu threads, rather than logical cpu threads

Slight improvement due to smallstr


jinyus:
Confirmed! It now matches the go con solution. Will merge and see how to performs on the workflow machine.

PR #39, opened 2023-09-30
Update run.sh

JalonSolov:
Simplest change to make the V version run faster.


jinyus:
Thanks, Oversight on my part. 5x boost on my machine.

PR #46, opened 2023-09-30
V: optimize `tagged_post_count` array to reset instead of allocating new

ArtemkaKun:
This PR optimizes allocations of `tagged_post_count` array to make it do the same (probably) what Go does under the hood.

### DIFF
--- a/v/related.v
+++ b/v/related.v
@@ -39,10 +39,10 @@ fn main() {
 	}
 
 	mut all_related_posts := []RelatedPosts{cap: posts.len}
+	mut tagged_post_count := []int{len: posts.len, cap: posts.len}
 
 	for i, post in posts {
-		// slower when preallocated
-		mut tagged_post_count := []int{len: posts.len, cap: posts.len}
+		unsafe { vmemset(tagged_post_count.data, 0, posts.len * tagged_post_count.element_size) }
 
 		for tag in post.tags {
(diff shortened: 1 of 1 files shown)


jinyus:
Hey, thanks for the PR but I'm not inclined to accept unsafe code as that opens Pandora's box for the other languages. I don't want to have a situation where others are submitting raw assembly to get ahead.

ArtemkaKun:
So if I introduce a PR into V language to implement `reset()` function for an array in std, that wouldn't be marked as `unsafe`, but will have exactly this implementation under the hood - this will be ok for you?

Because right now V code in this using a ton of `unsafe`, they are just hidden from the end user in the std. V requires `unsafe` block for every code that touches pointers, while Go, for example, is not.

The main disadvantage of V in this case is that the compiler is not smart enough to convert loop that doing zeros into `memset` call, like Go does, so it's need to be done manually.

jinyus:
The implementation is not the issue, I know go does the same. But I don't want to get into a situation where I have to review technically safe code written inside an unsafe block for other languages. We'll stick to language constructs for now.

ArtemkaKun:
Ok, I see, I will discuss that with the community and we will decide how to handle this not in the hacky way

jinyus:
Sounds good, I will close this PR for now.

PR #55, opened 2023-10-01
V reset array instead of allocate new

ArtemkaKun:
It turned out that in `-prod` C compiler already optimizes this to memset (I believe), so no unsafe shenanigans needed

### DIFF
--- a/v/related.v
+++ b/v/related.v
@@ -42,7 +42,9 @@ fn main() {
 	mut tagged_post_count := []int{len: posts.len, cap: posts.len}
 
 	for i, post in posts {
-		unsafe { vmemset(tagged_post_count.data, 0, posts.len * tagged_post_count.element_size) }
+		for j in 0 .. tagged_post_count.len {
+			tagged_post_count[j] = 0
+		}
 
 		for tag in post.tags {
 			for post_index in tag_map[tag] {
--- a/v/related.v
+++ b/v/related.v
@@ -39,7 +39,7 @@ fn main() {
 	}
 
 	mut all_related_posts := []RelatedPosts{cap: posts.len}
-	mut tagged_post_count := []int{len: posts.len, cap: posts.len}
+	mut tagged_post_count := []int{len: posts.len}
 
 	for i, post in posts {
 		for j in 0 .. tagged_post_count.len {
--- a/v/related.v
+++ b/v/related.v
@@ -39,10 +39,10 @@ fn main() {
 	}
 
 	mut all_related_posts := []RelatedPosts{cap: posts.len}
+	mut tagged_post_count := []int{len: posts.len, cap: posts.len}
 
 	for i, post in posts {
-		// slower when preallocated
-		mut tagged_post_count := []int{len: posts.len, cap: posts.len}
+		unsafe { vmemset(tagged_post_count.data, 0, posts.len * tagged_post_count.element_size) }
 
 		for tag in post.tags {
 			for post_index in tag_map[tag] {


spytheman (approved):
On my machine, with -prod, this loop has pretty much the same performance as using `unsafe { tagged_post_count.reset() }` .

spytheman:
Using this:
```v
...
import strings

fn json_encode(all_related_posts []RelatedPosts) string {
        mut sb := strings.new_builder(all_related_posts.len * 800)
        defer {
                unsafe { sb.free() }
        }
        sb.write_rune(`[`)
        for rp in all_related_posts {
                sb.write_rune(`{`)
                sb.write_string('"_id":"')
                sb.write_string(rp.id)
                sb.write_rune(`"`)
                if rp.tags.len > 0 {
                        sb.write_string(',"tags":[')
                        for t in rp.tags {
                                sb.write_rune(`"`)
                                sb.write_string(t)
                                sb.write_string('",')
                        }
                        sb.go_back(1)
                        sb.write_string(']')
                }
                if rp.related.len > 0 {
                        sb.write_string(',"related":[')
                        for p in rp.related {
                                sb.write_rune(`{`)
                                sb.write_string('"_id":"')
                                sb.write_string(p.id)
                                sb.write_string('","title":"')
                                sb.write_string(p.title)
                                sb.write_rune(`"`)
                                if p.tags.len > 0 {
                                        sb.write_string(',"tags":[')
                                        for t in p.tags {
                                                sb.write_rune(`"`)
                                                sb.write_string(t)
                                                sb.write_string('",')
                                        }
                                        sb.go_back(1)
                                        sb.write_string(']')
                                }
                                sb.write_string('},')
                        }
                        sb.go_back(1)
                        sb.write_string(']')
                }
                sb.write_string('},')
        }
        sb.go_back(1)
        sb.write_rune(`]`)
        return sb.str()
}

...
    // json_str_out := json.encode(all_related_posts)
    json_str_out := json_encode(all_related_posts)
...    

```

... improves the encoding performance *a lot* (from ~230ms to ~10ms on my machine).

With the loop and that custom json_encode function, the V program becomes pretty much equivalent to the ordinary Go program in performance (sometimes a bit faster, sometimes a bit slower).

I am going to improve the code generated for json.encode() soon, so that it will not have to be changed here (it is currently so slow, because it constructs a whole tree of AST json nodes for the structs, allocating memory for each AST node, then serialises that tree, instead of serialising the structs directly to a buffer).

jinyus:
This brought it up right up there with swift and crystal!

ArtemkaKun:
Yeah. We still don't understand why the timing is not the same as for go, because on dev machines it's the same

jinyus:
They are a little closer on my machine:

```
Go:

Benchmark 1: ./related
Processing time (w/o IO) 19.967851ms
Processing time (w/o IO) 23.219348ms
Processing time (w/o IO) 19.669974ms
Processing time (w/o IO) 19.425457ms
Processing time (w/o IO) 19.863995ms
Processing time (w/o IO) 20.10935ms
Processing time (w/o IO) 19.939495ms
Processing time (w/o IO) 21.679609ms
Processing time (w/o IO) 20.357639ms
Processing time (w/o IO) 21.814055ms
Processing time (w/o IO) 20.617242ms
Processing time (w/o IO) 21.599919ms
Processing time (w/o IO) 20.337036ms
  Time (mean ± σ):      42.3 ms ±   1.6 ms    [User: 38.3 ms, System: 12.5 ms]
  Range (min … max):    40.0 ms …  45.6 ms    10 runs
```

```
Vlang:

Benchmark 1: ./related
Processing time (w/o IO): 27.826ms
Processing time (w/o IO): 28.753ms
Processing time (w/o IO): 31.194ms
Processing time (w/o IO): 28.412ms
Processing time (w/o IO): 28.294ms
  Time (mean ± σ):     218.7 ms ±   6.8 ms    [User: 195.1 ms, System: 22.0 ms]
  Range (min … max):   210.5 ms … 228.2 ms    5 runs

```

PR #56, opened 2023-10-01
Python minor performance improvements and simplifications

fizmat:
1. report more detailed performance info (used [here](https://github.com/fizmat/related_post_gen_python_timing/blob/main/main.ipynb) to measure the rest of the changes)
2. run: install everything from requirements.txt (in case more libraries will be used in the future);
3. run: also remove venv in cleanup;
4. python: use defaultdict. Just convenience, no performance impact;
5. python: remove heapq. Counter can find top-k itself. Just convenience, no performance impact;
6. both: use ujson as a transparent replacement for json. 90 ms faster output;
7. numpy: generate unique tags using python set. 20 ms faster than filling an ndarray and then calling `.unique()`;
8. numpy: iterate over post.tags dict instead of iterating over all possible tag ids in the inner loop. 20ms faster;
9. use uint8 as matrix dtype: 20ms faster than uint16;

Interesting result in c9414160a56d5122a8ccb738a8675d394fdfa987: using `np.argpartition()` is 30ms slower than `np.argsort()`, even though partition does less (it does not have to fully sort the array). I think the key is `kind="stable"`. There's no algorithm selection for argpartition yet, and stablesort seems to be very fast on this specific dataset. Are related posts almost-sorted by relatedness?


fizmat:
Ooh, "stable" argsort is probably using radix sort in this case, which should be super fast given relatedness has very few possible values. https://numpy.org/doc/stable/reference/generated/numpy.argsort.html#numpy.argsort

jinyus:
Hey, thanks for your efforts but we need the timing to be compatible with the rest of the solutions because the benchmark results are automatically generated. The rest of your changes looks good.

Let me know if you're going to modify the pull request or I can do it.

fizmat:
Oh, I didn't know. What does the benchmark actually collect, the "Processing time (w/o IO): ..." lines?

jinyus:
Yes. I am also trying to limit the run time of the workflow because the free tier is limited.

[link to regex](https://github.com/jinyus/related_post_gen/blob/30c4b3067df562e35e0baa62372af90107e61ab3/extract_results.dart#L7)

fizmat:
Changed. Now it only prints the detailed table if env DETAILED_PYTHON_PERF="1", otherwise behaves as usual.
Also now using `time` functions that return int nanoseconds instead of float seconds (it will not actually improve precision given the current scale of ~ second per run, but why not).

jinyus:
Processing times is a bit slower on my machine. ujson sped up I/O though. What are the results on your machine?

new:

```
Benchmark 1: python3 ./related.py
Processing time (w/o IO): 1.425s
Processing time (w/o IO): 1.413s
Processing time (w/o IO): 1.421s
Processing time (w/o IO): 1.409s
Processing time (w/o IO): 1.415s
  Time (mean ± σ):      1.485 s ±  0.007 s    [User: 1.465 s, System: 0.018 s]
  Range (min … max):    1.478 s …  1.492 s    5 runs

```

old:

```Benchmark 1: python3 ./related.py
Processing time (w/o IO): 1.366s
Processing time (w/o IO): 1.353s
Processing time (w/o IO): 1.330s
Processing time (w/o IO): 1.349s
Processing time (w/o IO): 1.310s
  Time (mean ± σ):      1.516 s ±  0.023 s    [User: 1.492 s, System: 0.024 s]
  Range (min … max):    1.489 s …  1.546 s    5 runs
```

fizmat:
old:
```
Benchmark 1: python3 ./related.py
Processing time (w/o IO): 1.344s
Processing time (w/o IO): 1.351s
Processing time (w/o IO): 1.350s
Processing time (w/o IO): 1.342s
Processing time (w/o IO): 1.350s
  Time (mean ± σ):      1.497 s ±  0.005 s    [User: 1.477 s, System: 0.014 s]
  Range (min … max):    1.493 s …  1.503 s    5 runs
```

new:
```
Benchmark 1: python3 ./related.py
Processing time (w/o IO): 1.473s
Processing time (w/o IO): 1.465s
Processing time (w/o IO): 1.461s
Processing time (w/o IO): 1.458s
Processing time (w/o IO): 1.464s
  Time (mean ± σ):      1.508 s ±  0.027 s    [User: 1.486 s, System: 0.012 s]
  Range (min … max):    1.491 s …  1.555 s    5 runs
```

I checked, and now commit `minor pythonizations: deafultdict, no need for heapq` slows processing down by almost 120ms. Strange.

fizmat:
I think I know why I didn't see a difference before: version differences!

The latest result, where it's 120ms slower are in a py=3.9 venv, but the jupyter notebook with performance plots I generated in py=3.11. Probably either counter or defaultdict got optimized between 3.9 and 3.11.

jinyus:
I will merge and see how it performs in the VM

fizmat:
Yup, plot in 3.9:
image
Plot in 3.11:
image

jinyus:
I have 3.11.5 on my machine. It's slower on the vm too
[https://github.com/jinyus/related_post_gen/pull/59](https://github.com/jinyus/related_post_gen/pull/59)

jinyus:
I will have to revert for now.

fizmat:
The script does (a): `for p in heapq.nlargest(5, relation_count, key=relation_count.get)`
Internally the counter does (b): `for p, _n in heapq.nlargest(n, relation_count.items(), key=itemgetter(1))`
If I replace (a) with (b) in the script, it has the same performance as just calling relation_count.most_common(), which makes sense.

It seems (b) is measurably slower on almost all of my systems too. I wonder if it's because in (a) the heapq only stores the keys and gets counts from the counter, but in (b) the heapq stores the item-count pairs, so it uses 2x the memory.

Checked on a PC. (b) is slower than (a) on both Linux and Windows (both in python 3.11). Performance is the same *only* on M1 Mac in 3.11. I wonder how.

Win10 results (1: and 2: don't have the slower top5 from counter):
```
commit				import 		compile 	input 		process 		output 		total
0: install all pytho… 	11.90±0.18 	0.00±0 	10.42±0.26 	2744±18 	480.9±1.8 	3248±19
0: try counter most_… 	7.09±0.18 	0.00±0 	10.17±0.17 	2810±16 	54.28±0.99 	2882±16
1: minor pythonizati… 	12.08±0.28 	0.00±0 	10.27±0.26 	2729±13 	482.0±1.9 	3233±14
2: use ujson for IO…… 	7.12±0.23 	0.00±0 	10.21±0.17 	2718±14 	54.3±1.3 	2789±14
```
So this convenience change is definitely out.


By the way, on Windows the uint8 numpy matrix actually runs slower than uint16:
```
commit				import 		compile 	input 			process 		output 		total
0: minor pythonizati… 	115.92±0.72 	0.00±0 	10.35±0.21 		768.8±3.9 	481.6±1.5 	1376.7±4.0
1: use ujson for IO…… 	117.47±0.82 	0.00±0 	10.06±0.16 		770.8±3.4 	53.09±0.59 	951.4±4.0
2: np: use python se… 	118.21±0.91 	0.00±0 	10.031±0.098 	701.8±4.1 	52.74±0.63 	882.7±4.4
3: np: use smaller m… 	117.74±0.82 	0.00±0 	9.99±0.15 		743.1±2.1 	53.10±0.60 	923.9±2.7
4: np: use dict of t… 	118.3±1.3 	0.00±0 	10.03±0.14 		696.2±5.8 	53.15±0.61 	877.8±6.0
5: try to use numpy … 	117.56±0.87 	0.00±0 	9.95±0.17 		737.8±1.8 	53.03±0.68 	918.4±1.9
```

Is that a factor at all for you, or can performance differences on Windows be safely ignored?

PR #67, opened 2023-10-02
Node.js and Deno are not a "Language"

guest271314:
Node.js and Deno are JavaScript programming language runtimes, not languages in and of themselves.


jinyus:
Thanks

PR #68, opened 2023-10-02
add JAVA

venom-goblin:
Hello ! Here is a java version, it uses JAVA 17 and maven, I used java JMH instead of hyperfine to benchmark multiple runs as it is the standard solution in the ecosystem.

## Usage
Build and run the main program:
```bash
cd java
mvn -Pjvm clean package
java -jar target/main.jar
```

Build and run the JMH benchmarks:
```bash
cd java
mvn -PJMH clean package
java -jar target/benchmark.jar
```

Build and run the GraalVM program:
```bash
cd java
mvn clean package
# we execute the program a first time with an agent to collect metadata and generates the configuration files in a subdirectory of target/native/agent-output.
# Those files will be used by the native-image tool
mvn -Pnative -Dagent exec:exec@java-agent
mvn -Pnative -Dagent package
./target/related.exe
```

a GraalVM native profile is also available but not yet configured in the file .github/workflows/benchmark.yml

## Results:

On my machine (Windows) I had the following results:
run_go:
```
Processing time (w/o IO) 40.3919ms
106 ms
```

run_java:
```
Processing time (w/o IO) 180 ms
Processing (with IO) time 339 ms
```

run_java_with_jmh:
```
Processing time (w/o IO) 38 ms
Processing (with IO) time 65 ms
0,068 s/op
```

run java with graalvm:
```
Processing time (w/o IO) 61 ms
114 ms
```


jinyus:
Hey, thanks for the PR. I know nothing about java so bear with me, Why is benchmark.jar so much faster than main.jar? 
I'm more inclined to go with the real work tests (jit and graalvm) as those are running under the same conditions as the other languages.

Let me know what you think.

jinyus:
#50

venom-goblin:
benchmark.jar is faster because the JVM has a fixed startup cost and JMH reuses the same VM to run multiple iterations, and as the Java JVM uses a Just In Time compiler (JIT), the code runs first in interpreted mode and if a method is called enough times then it is eligible  for native compilation. The JIT code can even be faster than the Golang version because it has access to the runtime information of the machine and also the hot code paths executed.
So for server applications it can be beneficial but for short-lived command line ones it's less useful.

But I agree with you that it's good to have a neutral benchmarking tool for all languages.
We can just keep the run_java_with_jmh function in run.sh (but removing it from run_all) in case some people are interested about this scenario.

Thanks for merging the PR and adding support for GraalVM !

PR #73, opened 2023-10-02
Bun + Rust FFI (AI generated)

Airtune:


jinyus:
Hey, thanks for your efforts but FFI is not allowed as it's not testing the target language/runtime

PR #74, opened 2023-10-03
Speed up JS implementation

rawrmonstar:
`for..of` loops are usually faster than `Array.prototype` loops. I replaced the `Array.prototype` loops with `for..of` loops when you don't need to access the index.

On my computer for Node 20 it went from 124ms -> 85ms

https://github.com/jinyus/related_post_gen/assets/13814048/b516fd18-1dc5-4431-82af-35d1bdfba36a

### DIFF
--- a/js/related.js
+++ b/js/related.js
@@ -1,74 +1,65 @@
 
 export function genRelatedPosts(posts) {
-
   // build a map of tags to post indices. tag -> [idx1, idx2, ...]
-  const tagMap = posts.reduce(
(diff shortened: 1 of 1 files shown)


jinyus:
Thanks speed up for me too

jinyus:
It's surprising that this change resulted in a larger performance increase for Node and Deno compared to Bun. They are now 4x faster.

Also interesting that switching  `for in` to `forEach` sped up dart.

PR #75, opened 2023-10-03
add PGO for JAVA

venom-goblin:


jinyus:
Hey, how many % increase does this yield for you? 
Graalvm is already taking the most time to run in the workflow and the free tier is limited so I don't want it eating up my minutes. So I  might have to ignore this if it's not a significant speed up.

venom-goblin:
Hey !
The processing time without IO on my machine now takes 38ms instead of 60ms for GraalVM, and the current go version takes ~ the same time, but I understand the build time issue with Graal, so don't merge the PR yet I'll include the profiling file and the agent file in the PR to cut down build time.

jinyus:
Alright, sounds good

venom-goblin:
I included the profiling file and removed the need for the agent file.

PR #76, opened 2023-10-03
[IGNORE] Update benchmark results

jinyus:
New benchmark results


jinyus:
It's surprising that this change resulted in a larger performance increase for Node and Deno compared to Bun. They are now 4x faster.

Also interesting that switching for for in to forEach sped up dart.

PR #77, opened 2023-10-03
Add Nim

jinyus:
#54


michaelsbradleyjr:
Nim's `std/json` is well-known in the community to be on the slow side, and most projects will instead use `jsony` or another faster-than-std lib:

https://github.com/treeform/jsony

jinyus:
Added

michaelsbradleyjr:
> Added

Cool. So, there's some discussion in the Nim community about ways the performance of `related.nim` can be improved while adhering to the rules in this repo's readme and without getting exotic. Are you open to PRs in that regard?

jinyus:
Yes, go for it.

PR #80, opened 2023-10-03
Python (again)

fizmat:
Unlike the previous failed request (#56):
- only python, numpy will be separate
- keep manual use of heapq, as it's faster than counter.most_common (see #56)

P.S: speaking of numpy, on Linux and MacOS it's faster to use uint8 array, but on Windows uint16 array is faster.
1. Keep uint16, in case there is ever a post with more than 255 tags.
2. Use uint8, Linux is more important and 255 tags is enough
3. Check for platform at runtime. More complicated, more optimized (in theory). In practice would require more testing and retesting to make sure the choice stays optimal (for other architectures, etc.) Adding this basically creates an expectation of cross-platform optimization for the whole project that is very hard to meet.

I lean towards option 2, but ultimately it's your call.


jinyus:
2 sounds good to me

PR #81, opened 2023-10-03
Optimize Dart version

mraleph:
Main changes: 

* Dart's AOT compiler looses track of concrete types when containers-containing-other-containers are involved, e.g. `Map<K, List<T>>` or `List<List<T>>`. This makes operations wither nested containers slower. To work-around this problem you can simply wrap a nested container into some unique class - this helps AOT compiler to realise what comes from the container.
* Iterable.fillRange is slow - manually inlining it helps (I will file a bug against Dart to fix this).
* `for-in` is better than `forEach` when concrete types are properly known to the compiler (see the first item)
* Rewrote computation of top 5 to avoid unnecessary allocations

### DIFF
--- a/dart/related.dart
+++ b/dart/related.dart
@@ -1,73 +1,90 @@
 import 'dart:convert';
 import 'dart:io';
+import 'dart:typed_data';
 
 import 'models.dart';
 
+// When you have type `Map<String, List<int>>` Dart AOT compiler can't figure
+// out that this container can contain only specific type of lists as
+// values. However if you create an indirection (like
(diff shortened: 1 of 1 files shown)


mraleph:
These changes improve Dart AOT performance from 183ms to 100ms on my machine (measured on HEAD of Dart SDK - the upcoming release has a bunch of improvements which make `for-in` perform better in AOT mode).

jinyus:
Thanks for the PR. Very insightful. Drops from 70ms -> 55 on my machine.

PR #88, opened 2023-10-04
Fixed in/out paths for fsharp

JordanMarr:
Fixed in/out path to properly find and save json files in the root dir.


jinyus:
Does having forward slashes (/) work on windows?

JordanMarr:
Yes, forward and backslashes both work on Windows.

PR #90, opened 2023-10-04
Python: use scipy sparse array over numpy dense array

fizmat:
As not every post has every tag, the post-tag matrix is sparse (has many 0s). Matrix multiplication is faster if we rely on this.
1. moving from numpy to scipy processing jumps from 500-700 ms to 270-280 ms;
2. using internals of sparse array representation (Compressed Space Row), sorting for top5 wins additional 40ms;

commit | import | compile | input | process | output | total
-- | -- | -- | -- | -- | -- | --
0: copy numpy matrix… | 47.52±0.88 | 0.00±0 | 3.548±0.023 | 476.7±4.3 | 18.25±0.98 | 546.1±5.8
1: use scipy sparse … | 81.6±5.2 | 0.00±0 | 3.571±0.082 | 276.0±3.9 | 18.8±2.0 | 380.0±5.8
2: sort results dire… | 78.2±1.2 | 0.00±0 | 3.530±0.035 | 228.4±7.8 | 18.5±1.1 | 328.6±8.4

Should this stay a separate implementation (and have a new run.sh entry), or should it replace the Numpy implementation? The logical idea is the same but faster, but with more dependencies (even import takes extra 30ms).

### DIFF
--- a/python/related_np.py
+++ b/python/related_np.py
@@ -3,6 +3,7 @@ from timing import lap, finish
 lap()
 import ujson as json
 import numpy as np
+from scipy.sparse import coo_array
 
 
 def main():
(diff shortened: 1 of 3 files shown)


jinyus:
Replace the numpy version

fizmat:
Done. Also fixed a bug in the second commit.

PR #91, opened 2023-10-04
Improve Nim solution

PMunch:
Slightly better code, mostly low-hanging fruit, could definitely be optimised further. Also added LTO and PGO to the compilation which is common when going for maximum speed. Could compile with `-d:danger` for even more speed but that technically removes some bound checks and violates one of the rules. I guess `-d:danger --boundChecks:on` would still be allowed though, would give a slight increase.

### DIFF
--- a/nim/src/related.nim
+++ b/nim/src/related.nim
@@ -1,46 +1,38 @@
 # This is just an example to get you started. A typical binary package
 # uses this file as the main entry point of the application.
-import std/tables
-import std/sequtils
-import times
+import std/[tables, sequtils, monotimes, times]
 import jsony
 
 type
-  Post = object
+  Post = ref object
     `"_id"`: string
     title: string
     tags : seq[string]
 
-  RelatedPosts = object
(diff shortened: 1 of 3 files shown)


PMunch:
By the way, these changes about halves the time on my machine.

nim/src/related.nim
```diff
@@ -1,81 +1,66 @@
...
-
-for i, post in posts.pairs:
+var tagMap: Table[string, seq[int]]
+for i, post in posts:
   for tag in post.tags:
-    var lst = tagMap.getOrDefault(tag, @[])
-    lst.add(i)
-    tagMap[tag] = lst
+    if tagMap.hasKeyOrPut(tag, @[i]):
+      tagMap[tag].add i
```
Menduist, line 29:
```suggestion
    tagMap.mgetOrPut(tag, @[]).add(i)
```

Your version is still doing a double lookup on happy case, I'd expect a single lookup to slightly improve things

PR #95, opened 2023-10-04
Add basic Swift concurrent version

yaglo:
Same code inside as in the normal version, but parallelized with Dispatch and uses `TreeDictionary` from <https://github.com/apple/swift-collections>.

When I have a free minute, I'll look into the actual algorithm.


jinyus:
Thanks, I will have to re-enable bounds checking as removing it is against the [rules](https://github.com/jinyus/related_post_gen#no). Good news is that it doesn't affect the score on my machine

PR #96, opened 2023-10-04
Fsharp parallel

matthewcrews:
- Used faster collections for gathering the data for Tag-Post relationships
- Switched to System.Diagnostics.Stopwatch for timing for accuracy
- Parallelized the creation of the result array
- Use stackalloc for collecting intermediate results
- Updated directory input/output to work on Windows as well

### DIFF
--- a/fsharp/Program.fs
+++ b/fsharp/Program.fs
@@ -1,7 +1,14 @@
 ﻿open System
 open System.IO
+open FSharp.NativeInterop
 open FSharp.Json
 open System.Collections.Generic
+
+#nowarn "9"
+
+let inline stackalloc<'a when 'a: unmanaged> (length: int): Span<'a> =
+  let p = NativePtr.stackalloc<'a> length |> NativePtr.toVoidPtr
+  Span<'a>(p, length)
 
 
 [<Struct>]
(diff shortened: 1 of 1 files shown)


jinyus:
Did you measure the postsStack. It's slower than accessing the tags on the struct on my machine:

new

```
for tagId in postTagsStacks[postId] do
```

```
Processing time (w/o IO): 57ms
Processing time (w/o IO): 61ms
Processing time (w/o IO): 52ms
Processing time (w/o IO): 49ms
Processing time (w/o IO): 59ms
Processing time (w/o IO): 47ms
Processing time (w/o IO): 55ms
```

old:

```
for tagId in post.tags do
```

```
Processing time (w/o IO): 47ms
Processing time (w/o IO): 48ms
Processing time (w/o IO): 47ms
Processing time (w/o IO): 49ms
Processing time (w/o IO): 47ms
Processing time (w/o IO): 48ms
Processing time (w/o IO): 50ms
```

matthewcrews:
@jinyus I'm seeing the same thing. I've updated the PR.

matthewcrews:
@jinyus Are you open to custom versions of data structures? I have a faster version of Dictionary for `string` keys that I use. I'd include the code.

matthewcrews:
@jinyus Let's just stick with this version. I don't think bringing in custom data structures is in the heart of this comparison 😊

jinyus:
We don't even need a map, I experimented with a [version](https://github.com/jinyus/related_post_gen/commit/aff200fe70220d5ed9a2cdf2665abdc347c1f75c) that used ints to represent tags and used a regular array instead of a map but it only yield microsecond gains so I dropped it as every other language would need to be updated and it's not worth it.

matthewcrews:
I tried the same thing and didn't see a performance improvement so I dropped it.

jinyus:
I need to create separate version for single thread and parallel but for some strange reason, when I use `Array.mapi` instead of `Array.Parallel.mapi`, isn't slower than the original and when I parallelize the original is slower than the new version.

matthewcrews:
I created a separate PR where I showed the single-threaded version. I closed the PR because it will have conflicts with this one. You can see what I did though.

PR #99, opened 2023-10-04
FSharp: use int[] instead of Stack<int> for faster access

pkese:
Turns out that dotnet runtime generates faster code for accessing `array<int>` than `Stack<int>`


pkese:
Note:  
I dared to rename the project from `fsharp.fsproj` to `fsharp_con.fsproj` to make it easier to find my way around in source code editor.  
I hope that is ok.

jinyus:
Very insightful, this halves the time on my device.

PR #106, opened 2023-10-05
Assorted improvements to Rust benchmarks

mcronce:
In no particular order

* Bring back `mimalloc` in both cases
* Bring back `smallstr` for ID and tags; it's faster than borrowing from the original JSON blob, presumably due to better cache locality
  * Also port that optimization to `rust` (previously it was only used in `rust_con`)
  * Once `compact_str` gets another release, it will provide a significant performance improvement over `smallstr`, but I don't expect pulling a dependency from github instead of crates.io is something you want
* Actually borrow titles
* Use `u16`s instead of `usize`s for post index and count (someone changed them to `u32`s earlier today)
* Use slices for references to strings and vecs
* Use thin LTO for `rust_con`; explicitly use fat for `rust`
* Mark trivial functions (`eq()`, `partial_cmp()`, and `cmp()` on `PostCount`) with `#[inline]` markers; this makes them eligible for cross-crate inlining, which is relevant when they're called from e.g. the std library

Each of these made a small but repeatable change, at least on my local machine

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -183,6 +183,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9cdc71e17332e86d2e1d38c1f99edcb6288ee11b815fb1a4b049eaa2114d369b"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3979b5c37ece694f1f5e51e7ecc871fdb0f517ed04ee45f88d15d6d553cb9664"
+dependencies = [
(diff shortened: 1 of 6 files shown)


Dr-Emann:
I feel like u16 might be a bit too far out of realism, 64k posts isn't _that_ much.

mcronce:
🤷‍♂️ there are 5k in the benchmark.  Effective optimization includes exploiting domain specific knowledge.

jinyus:
I will add a requirement that it should handle 100k posts because I'm going to remove the total column and show how the program scale for a varying number of posts. I will manually update the codes so you don't have any extra work. But you can submit a PR if you wish.

PR #109, opened 2023-10-05
use flattened list for nim

jinyus:


PMunch:
Is this actually any faster? Tuples in Nim are value types, so the memory layout of this should be the same as the other version.

PR #111, opened 2023-10-05
Further speedup of Nim version

PMunch:
This has three improvements:
- Uses the FxHash algorithm which is used in the Rust solution
- Performs a marginally faster zeroMem operation instead of newSeq.
- Packs the counts tighter together for faster looping by using a uint8 type

### DIFF
--- a/nim/src/related.nim
+++ b/nim/src/related.nim
@@ -1,6 +1,4 @@
-# This is just an example to get you started. A typical binary package
-# uses this file as the main entry point of the application.
-import std/[tables, sequtils, monotimes, times]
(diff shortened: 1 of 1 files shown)


jinyus:
Thanks, this has some merge conflicts so it will take some time to get it merged. It's no longer using the tuple, I got some speed up by using a flattened list. Would be great if you could roll the new changes into your version to resolve the conflict.

PMunch:
Yes I noticed as I created the PR. Will definitely have a look at the flattened list thing, that shouldn't really make a difference so it will be interesting to check anyways.

jinyus:
I'm doing some tests and the flatten list approach is now slower. I will merge this.

PR #117, opened 2023-10-05
Increase speed a little

DrBlury:


jinyus:
Hey, the number of posts has to be dynamic and it needs to show an error if decoding fails. Every other solution does this.

PR #119, opened 2023-10-05
Make F# code a bit more idiomatic

pkese:
I've tried to clean up F# code a little bit (just the non-concurrent version).  
It is slightly more allocation heavy, but it also measures marginally faster on my machine.

May I ask you, if you could benchmark it, before applying this pull request.


jinyus:
Maginally faster on my machine, thanks.

PR #120, opened 2023-10-05
Added cpp example, pls update cpp devs

DrBlury:


jinyus:
Nice starting point, it takes 4s for me so I'm not going to included it in the chart until it mirrors another impl.

PR #123, opened 2023-10-05
Julia Naive Fix

algunion:
I feel that the previous implementations were not entirely fair to Julia's language, so I did a kind of a _naive_ fix here.

I kept the algorithms straightforward without using sophisticated (or unfair) Julia performance tricks.

A concurrent version will follow.


jinyus:
Nice, 30x improvement for me

PR #128, opened 2023-10-05
Use top-level statements in the C# implementation

aradalvand:
This is more of a stylistic change, but you actually didn't need to have a `Program` class with a `Main` method, it's redundant code and adds two unnecessary levels of indentation — C# 10 introduced top-level statements, eliminating the need for explicitly declaring these. The only thing is that the types need to come after the statements.


jinyus:
Thanks

PR #130, opened 2023-10-05
[SKIPCI] Update benchmark results

jinyus:
New benchmark results

### DIFF
--- a/raw_results.md
+++ b/raw_results.md
@@ -1,394 +1,394 @@
 Go:
 
 	Benchmark 1: ./related
-	Processing time (w/o IO): 30.19699ms
-	Processing time (w/o IO): 30.477189ms
-	Processing time (w/o IO): 30.676989ms
-	Processing time (w/o IO): 30.04849ms
-	Processing time (w/o IO): 30.414789ms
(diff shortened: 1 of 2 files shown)


PallHaraldsson:
The ranking doesn't seem reliable, run on same machine, now loaded? Fastest language to 2nd fastest, but old time would still keep it at the top.

Also similar story with new:

Go Concurrent | 18.89 ms | 49.6 ms |
Rust Concurrent

Also Julia is missing there, but same non-concurrent would do well on that list.

jinyus:
It's ran in a github workflow, I guess a different instance is provisioned every run. Even running on my machine doesn't get the same results everytime. I think I  would need to rent a dedicated server to get consistent results.

Julia is 5th in the rankings.

PallHaraldsson:
When I test I turn of the web browser. Ranking should at least be stable over time on the same machine.

If it's a VM and not, then it's likely one of many on same physical hardware. If you can't do locally on your own machine I would at least run more than once and see if the ranking doesn't change. The numbers will never be exactly the same, you could get same ranking though, or take minimum across runs.

PR #132, opened 2023-10-06
Performance Improvement (Julia)

algunion:
Some minor changes - performance improvement related.


algunion:
Let me check the correctness before checking this.

jinyus:
This version is not correct, you need to modify the run.sh to check `related_posts_julia.json`. It's currently checking `related_posts_julia_v1.json`

algunion:
> This version is not correct, you need to modify the run.sh to check `related_posts_julia.json`. It's currently checking `related_posts_julia_v1.json`

Yes - I figured that out right away.

algunion:
@jinyus - please check now.

jinyus:
I'm seeing 1-2ms speed up

algunion:
> I'm seeing 1-2ms speed up

I'll need to go over it in more depth later - I'll leave it in this state for now. 

The goal was to bring the stats closer to Rust (because values > 0.5 seconds were pretty weird for Julia).

Also - you did an excellent job with a real-world benchmark (as opposed to useless micro-benchmarks).

jinyus:
Thanks, this bench mark was formed from a real world problem. 

I was using golang for a similar problem and it was too slow so I tried rust and it was even slower. That's when I posted my findings to the rust community and I thought this be a great test for other languages

PR #136, opened 2023-10-06
FSharp Reuse top5 array

matthewcrews:
- Reuse the top5 array
- Reduce the size of taggedPostCount array using bytes since the number of tags is <=100

This gave me a small speed boost on my machine from ~20ms to ~17ms
I'm running an Intel 13900K


jinyus:
shaves a couple ms for me too

PR #139, opened 2023-10-06
CSharp Replace stack with list and struct with class

NightlyRevenger:
- Replace struct with class
- Replace Stack with List

This gave me 30% speed boost on my machine i5-10400T


jinyus:
Hey, myself and at least 4 others have confirmed that the stack -> array version is faster.

[https://github.com/jinyus/related_post_gen/pull/140#issue-1929603092](https://github.com/jinyus/related_post_gen/pull/140#issue-1929603092)

PR #140, opened 2023-10-06
C# optimizations

zigzag312:
- AOT compilation.
- ~tagMap initialization without copying collections again.~ Surprisingly, pushing to stack and then converting to array is faster than using List<int>
- Reduced dictionary lookups in the loop to single lookup.
- Removed required keyword (AOT JSON serialization doesn't build in .NET 7 with `required` keyword, this might be fixed in .NET 8). 
- changed structs to classes as structs are copy by value by default (improves performance in this case)

### DIFF
--- a/csharp/related.csproj
+++ b/csharp/related.csproj
@@ -5,6 +5,7 @@
     <TargetFramework>net7.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <PublishAot>true</PublishAot>
   </PropertyGroup>
 
 </Project>
--- a/csharp/Program.cs
+++ b/csharp/Program.cs
@@ -96,7 +96,7 @@ Console.WriteLine("Processing time (w/o IO): {0}ms", sw.Elapsed.TotalMillisecond
 
 File.WriteAllText(@"../related_posts_csharp.json", JsonSerializer.Serialize(allRelatedPosts, MyJsonContext.Default.RelatedPostsArray));
 
-public struct Post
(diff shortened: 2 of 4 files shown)


csharp/related.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>net7.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <PublishAot>true</PublishAot>
```
Szer, line 8: This flag requires explicit RID on publish like
`dotnet publish -r linux-arm64 -c Release`

Also please make sure both build scripts are working correctly and executables paths are correct
https://github.com/jinyus/related_post_gen/blob/main/run.ps1#L81-L93

https://github.com/jinyus/related_post_gen/blob/9db7a779bc431ec5b0d23e25e958bdf3eedd0128/run.sh#L374-L386

zigzag312: I didn't modify build scripts as I only tested building from VS on Windows. I'm not familiar with hyperfine, but I'll try dotnet publish command...

zigzag312: It builds but output folder of AOT build is changed

zigzag312: I added output paths to scripts. I think it should work now.

neon-sunset:
> Removed required keyword (AOT JSON serialization doesn't build in .NET 7 with required keyword, this might be fixed in .NET 8).
You just need to reference preview .NET 8 serializer, you can keep TFM itself at .NET 7.

PR #144, opened 2023-10-06
CSharp - Optimize Dictionaries / Enable Server GC

mnj:
Do like Go, assume less than 100 elements.


mnj:
Each of these two small fixes shaved about 5-8ms off each on my PC (more when run on the Github Runner).

PR #146, opened 2023-10-06
Perf Improvement - StrideArrays (Julia)

algunion:
I just ran a fast experiment using `StrideArrays` - it seems to improve the GC work and has alt least 2ms impact performance on my machine.

### DIFF
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -2,6 +2,7 @@ using JSON3
 using StructTypes
 using Dates
 using StaticArrays
+using StrideArrays
 
 function relatedIO()
     json_string = read("../posts.json", String)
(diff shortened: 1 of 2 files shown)


algunion:
Someone on Julia discourse pointed out that using StrideArrays breaks the rules (because it seems the StrideArrays implementation does not check bounds by default.

This seems to be, indeed, the case - their Github repo showcases a warning about this (but the documentation does not mention anything about the subject).

Now that I know, it is only fair to add the disclaimer here - and also apologize for not digging deep enough at the time.

I'll do an update soon, keeping the _legal_ stuff from this PR and removing/updating the stuff around StrideArrays.

PR #148, opened 2023-10-06
String -> Symbol for tags

algunion:
Since we can use the information that we have a maximum of 100 tags, Julia seems to behave better if (in the context) we store the tags as `Symbol` (instead of `String`).

I noticed that it reduces the variability of the runs and also gained about 1-2 ms on average on my machine.

### DIFF
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -25,12 +25,12 @@ end
 struct PostData
     _id::String
     title::String
-    tags::Vector{String}
+    tags::Vector{Symbol}
 end
 
 struct RelatedPost
     _id::String
-    tags::Vector{String}
+    tags::Vector{Symbol}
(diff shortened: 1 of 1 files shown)


jinyus:
A few solutions did this but I said no. They wanted to represent tags as int. Let keep the hashing involved.

algunion:
> A few solutions did this but I said no. They wanted to represent tags as int. Let keep the hashing involved.

I understand, and I think is fair to make sure we compare apples to apples here.

I didn't check if other languages did this - but please keep in mind that a Julia `Symbol` is basically an [interned string](https://en.wikipedia.org/wiki/String_interning). So, hashing would still be involved.

However, I am OK with going with your decision here as long as you act as a steward and enforce the same rules for all the other languages.

PR #149, opened 2023-10-06
Csharp custom json serialization

MarkPflug:
Significantly reduces the IO time (not the processing time) by using hand-rolled JSON serialization.
Maybe this is a step too far, and violates the spirit of the comparison, but when your time is dominated by serialization, it doesn't make much sense to optimize the data processing code.


jinyus:
> Maybe this is a step too far, and violates the spirit of the comparison

You're right, this would definitely be useful in a certain context. But the benefit is not exclusive to c#

PR #150, opened 2023-10-06
FSharp Unroll loop

matthewcrews:
- Unroll the loop for totaling tag count overlaps.
- I observed ~15% speed up

### DIFF
--- a/fsharp/Program.fs
+++ b/fsharp/Program.fs
@@ -59,8 +59,19 @@ let allRelatedPosts: RelatedPosts[] =
         let taggedPostCount = stackalloc posts.Length
 
         for tagId in post.tags do
-            for relatedPostId in tagPosts[tagId] do
-                taggedPostCount[relatedPostId] <- taggedPostCount[relatedPostId] + 1uy
+            let relatedPostIds = tagPosts[tagId]
+            let mutable i = 0
+
+            while i + 3 < relatedPostIds.Length do
+                taggedPostCount[i  ] <- taggedPostCount[i  ] + 1uy
+                taggedPostCount[i+1] <- taggedPostCount[i+1] + 1uy
+                taggedPostCount[i+2] <- taggedPostCount[i+2] + 1uy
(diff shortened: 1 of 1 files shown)


jinyus:
Nice hack but this makes the code very unreadable and I don't want everybody start doing this. Going to keep this in mind for future references, thanks.

matthewcrews:
Um, this is a very common optimization technique. You'll find this kind of thing often in high-performance code. Do you want me to comment on it more so it's clearer what is happening?

matthewcrews:
I mean, [Loop Unrolling](https://en.wikipedia.org/wiki/Loop_unrolling) is a well-documented technique, and other languages are not precluded from using it.

If the benchmarking intends to show how fast a language can go, I would use this technique. If the goal is to show the performance of idiomatic code, I would say that even the initial version with the custom Priority Queue would not be accepted.

I guess I'm confused about what the rules are for what is and is not acceptable 😕. A large part of my job is writing fast F# code and I have used this technique several times.

matthewcrews:
Would you be open to having an F# Normal version and an F# Optimized version?

A "Normal" F# version would not use `stackalloc` or the custom PriorityQueue logic, nor would it reuse any arrays. This would be the simple, obvious approach.

An F# optimized version would use all the techniques we've been integrating.

The reason I'm commenting is that people often look at these benchmarks and say, "Well, this language is slow!" when, in reality, they are only seeing one approach to the problem. As an advocate for the F# language, I like to show people how far they can push the language, so I invest time in showing people how to push the performance of F#. I have an entire YouTube channel for it: [Fast F#](https://www.youtube.com/channel/UCnOx0OGml1t-6u_v2N0GemA)

jinyus:
You have good points. I should've require readable, idiomatic code from the start but I guest that ship has sailed.

matthewcrews:
No worries.

Also, my apologies; there was a bug in my code that caused the performance to be better than it really is, so I deleted the branch.

This has been fun and thank you for hosting this 😊

PR #153, opened 2023-10-06
Numba (Python JIT compiler, a little cheating)

fizmat:
- a numpy solution not based on matrix multiplication, but adding arrays in a loop. Slower than the matrix solution, on par with the original Python solution: 1.5s processing, 1.5s total
- JIT compile the inner loop function with Numba: instantly overtake even scipy solution on processing: 140 ms processing, 650ms total (because of 180ms import and 310ms JIT compile)
- sorting to find top5 takes almost half the processing time. replace sorting with one pass over the array to save 55ms (*this part probably can be optimized*).
- other minor optimizations. Final: 60ms processing, 1.2s total (mostly compile)
- parallelize the outer loop: 27ms processing, 1.6s total (even longer compile)

Should the parallelized solution be a separate entry in run.sh?

It's not hard to use just one python script for both sequential and parallelized solution, but probably simpler and more consistent with the other solutions to have a different .py file for the parallel version.

Performance measurments: https://github.com/fizmat/related_post_gen_python_timing/blob/main/main.ipynb

### DIFF
--- a/python/related_numba.py
+++ b/python/related_numba.py
@@ -32,6 +32,16 @@ def get_top5(relation_count):
     return max_i
 
 
+@njit
+def get_all_top5(n_posts, t_to_pp, p_to_tt):
+    all_top5 = np.empty((n_posts, 5), np.uint16)
+    for p in range(n_posts):
+        relation_count = count_relations(n_posts, t_to_pp, p_to_tt, p)
+        top5 = get_top5(relation_count)
+        all_top5[p] = top5
+    return all_top5
+
+
 def precompile():
     lap()
     # JIT compile by running with the arguments of the correct type
@@ -43,6 +53,11 @@ def precompile():
         typed.List([np.empty(0, dtype=np.uint8)]),
         0)
     get_top5(np.empty(0, dtype=np.uint8))
+    get_all_top5(
+        0,
+        typed.List([np.empty(0, dtype=np.uint16)]),
+        typed.List([np.empty(0, dtype=np.uint8)])
+    )
 
 
 def main():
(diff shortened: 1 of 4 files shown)


jinyus:
create another entry in run.sh for concurrent python, follow the naming of other concurrent solutions

fizmat:
Done. I hope the naming is correct.

python/related_numba.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def main():
+    lap()
+    with open("../posts.json", "rb") as f:
+        s = f.read()
+        posts = orjson.loads(s)
+    lap()
+
+    unique_tags = set(tag for post in posts for tag in post["tags"])
```
jinyus, line 56: tags have to be represented as strings. Check the [rules](https://github.com/jinyus/related_post_gen#updated-results-from-github-workflow-raw-data)

fizmat: 1. This change did not start in this PR. dd1f128be1e706aa9e28cefe5ab0619ec3894ca5
2. What is actually the problem? This is a set of strings.

jinyus: I meant this line:

```
tag_to_t = {t: np.uint8(i) for i, t in enumerate(unique_tags)}
```

fizmat: Impossible to use strings as array indices. Not in numba, not in numpy, not in C or Rust. Are you sure the rules are this strict: no numbering tags in any way?
Logically it's the same as having a matrix of size (tag_n x post_n) that has tag numbers as matrix indices. You will have to delete the numpy solution too. There is no workaround.

fizmat: Don't get me wrong, I will feel bad if both solutions are deleted, but if the aim of the repo is to benchmark just one exact algorithm written in different languages, then so be it. I think trying to trick slow languages into solving the problem quickly is fun, but this is not actually a clever solution competition and the rules not mine to write.

jinyus: The rule was made because everyone wanted to optimize away hashing and just use arrays. I think I will review the code and decide.

fizmat: I wonder if there's a way to tweak the problem so that hashing is actually optimal.

jinyus:
Yea, I will merge but I can't put it in the charts as it's not fair to the other langs.

PR #154, opened 2023-10-06
C# Using Spans for better memory allocation times

AJHopper:
most of the time seems spent in stack pushes, could potentially get that down but would be some weird code

### DIFF
--- a/csharp/Program.cs
+++ b/csharp/Program.cs
@@ -4,13 +4,15 @@ using System.Runtime.InteropServices;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
+
 const int topN = 5;
-var posts = JsonSerializer.Deserialize<List<Post>>(File.ReadAllText(@"../posts.json"), MyJsonContext.Default.ListPost);
+var posts = JsonSerializer.Deserialize<List<Post>>(File.ReadAllText(@"../posts.json"), MyJsonContext.Default.ListPost)!;
+var postsCount = posts.Count;
 
 var sw = Stopwatch.StartNew();
 
 // slower when int[] is used
-var tagMapTemp = new Dictionary<string, Stack<int>>(100);
+var tagMapTemp = new Dictionary<string, LinkedList<int>>(100);
 
 for (var i = 0; i < posts!.Count; i++)
 {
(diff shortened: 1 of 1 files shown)


csharp/Program.cs
```diff
@@ -4,38 +4,34 @@
...
 {
     foreach (var tag in posts[i].Tags)
     {
         // single lookup
-        ref var stack = ref CollectionsMarshal.GetValueRefOrAddDefault(tagMapTemp, tag, out _);
-        stack ??= new Stack<int>();
-        stack.Push(i);
+        ref var stack = ref CollectionsMarshal.GetValueRefOrAddDefault(tagMap, tag, out _);
+        stack ??= new int[100];
+        _ = stack.Append(i);
```
zigzag312: This looks like a bug. Append is extension method that creates a new IEnumerable instance and you are throwing the result away while the source array stays unmodified.

jinyus:
Hey, you could get this merged faster by resolving the conflicts. If not, it will have to wait a few hours.

zigzag312 (approved):

csharp/Program.cs
```diff
@@ -46,40 +50,38 @@
...
-            minTags = top5[topN * 2 - 2];
+            minTags = top5[topN - 1].Count;
         }
     }
 
     var topPosts = new Post[topN];
 
     // Convert indexes back to Post references. skip even indexes
-    for (int j = 1; j < 10; j += 2)
+    for (int j = 0; j < 5; j ++)
```
zigzag312, line 82: Good catch. I forgot to change staring index in my PR.

csharp/Program.cs
jinyus: That version is outdated

AJHopper: Yes, was a bug - ended up reverting back to the Stack that was being used - swapped it out for a LinkedList instead of a stack though in the end cause, LL is very, very slightly faster

AJHopper:
> Hey, you could get this merged faster by resolving the conflicts. If not, it will have to wait a few hours.

Should be merged and updated, test run locally an seems to be all good

csharp/Program.cs
zigzag312: I tried LinkedList and it's slightly slower on my machine. List and Stack seem to be the same. The most impact has the conversion to array at the end (tagMapTemp->tagMap).

AJHopper: Interesting, LinkedList should add things faster that a Stack will, but access / traversal is generally slower. For me it was consistently about 4ms faster on my machine. using LinkedList, but could easily be possible on different hardware etc it swings the other way

zigzag312: LinkedList.AddLast is constant time, but has a larger overhead for each call. List is usually faster when adding items at the end (same for Stack.Push and Queue.Enqueue).

PR #157, opened 2023-10-08
return valid results from genRelatedPosts in js/related.js

michaelsbradleyjr:
`Array(posts.length)..fill({})` results in the same object in `.map((record, i) => {...}` being used in each iteration, which results in the array returned by `genRelatedPosts` being invalid. The solution is simple: use a unique `record` object in each iteration.

The problem (undetected by `verify.py`) was found with the help of `verify.sh` and `verify.jq`, which are introduced by this PR. The new verification makes use of [`jd`](https://github.com/josephburnett/jd) and is fast, taking less than a second on my laptop per `./verify.sh related_posts_[name].json`. The GitHub Actions workflow is updated to install `jd` (`${HOME}/go/bin` is already in `$PATH` in that environment).

On some platforms the `time` command is BSD not GNU, and the latter may be installed as `gtime`; so in `run.sh` define and use a `${time}` variable that accounts for that: helpful for those of us doing `./run.sh ...` on e.g. macOS. Similarly, `/bin/bash` is an ancient version on macOS, so instead of `#!/bin/bash` specify `#!/usr/bin/env bash` which can resolve to a modern version e.g. installed as `/usr/local/bin/bash`.

The `correct_related.json` in this PR contains the same data as before, but has been output by `jq` in a more compact format.


jinyus:
Hey, nice catch and thanks for your efforts. I must that I'm reluctant to add another dependency for users when the bug can be fixed by updating the current script [https://github.com/jinyus/related_post_gen/commit/e74b7379f4f81c91b5d26230e73bafb109e4e142](https://github.com/jinyus/related_post_gen/commit/e74b7379f4f81c91b5d26230e73bafb109e4e142). I don't want to put any hurdles up for someone who just wants to test/add a single language. I also don't understand the jq syntax and it's not something I can commit the time to learn right now. 

The rest of you changes look good so I will merge and just remove the extra files manually so I don't give you extra work.

PR #163, opened 2023-10-08
[julia] put Julia code into a project better precompile

Moelf:
This should significantly reduce the "total" time, which before was unfairly showing the compilation time.

The native code caching was introduced in 1.9: https://julialang.org/blog/2023/04/julia-1.9-highlights/#caching_of_native_code

## this PR

```
  Time (mean ± σ):     407.2 ms ±  12.5 ms    [User: 430.0 ms, System: 338.5 ms]
  Range (min … max):   394.0 ms … 431.2 ms    10 runs

Checking output
Success: related_posts_julia.json is valid!
```


jinyus:
Hey, I'm going to remove the total time column and show how the program scales with different number of posts. So I think this isn't necessary. Let me know if you object. 

| Language       | 5k posts | 20k posts | 100k posts |
| -------------- | --------------- | ---------------- | ---------------- |

Moelf:
as long as we aren't using the time of invoking `julia related.jl` until it return to shell, I'm fine.

That time includes start up Julia, import packages, and compile the code. (because it's not in a package these native code cannot be cached_

PR #166, opened 2023-10-08
go: reduce GC using a soft memory limit

felixge:
This PR reduces the GC of the Go implementations by setting a soft memory limit that is high enough to avoid the GC kicking in. This speeds up the JSON decoding and encoding at the beginning and end. It doesn't speed up the processing loop in the middle.

Below are the results from my M1 MBP, YMMV.

## ./run.sh go

<table>
<tr>
	<td>
Before

CleanShot 2023-10-08 at 23 04 11@2x

```
  Time (mean ± σ):      39.7 ms ±   1.7 ms    [User: 33.9 ms, System: 11.2 ms]
  Range (min … max):    35.9 ms …  41.6 ms    10 runs
```
</td>
</tr>
<tr>
	<td>
After

CleanShot 2023-10-08 at 23 03 15@2x

```
  Time (mean ± σ):      35.1 ms ±   0.6 ms    [User: 28.0 ms, System: 5.4 ms]
  Range (min … max):    33.8 ms …  36.2 ms    10 runs
```
</td>
</tr>
</table> 

## ./run.sh go_con

<table>
<tr>
	<td>
Before

CleanShot 2023-10-08 at 23 24 24@2x

```
  Time (mean ± σ):      25.3 ms ±   1.9 ms    [User: 38.0 ms, System: 11.1 ms]
  Range (min … max):    22.9 ms …  28.0 ms    10 runs
```
</td>
</tr>
<tr>
	<td>
After

CleanShot 2023-10-08 at 23 28 39@2x

```
  Time (mean ± σ):      20.9 ms ±   0.5 ms    [User: 31.9 ms, System: 7.4 ms]
  Range (min … max):    20.3 ms …  22.0 ms    10 runs
```
</td>
</tr>
</table> 

## more ideas

For amd64 I'm wondering if setting `GOAMD64=v4` during the build would squeeze out some extra performance. I didn't have time to try on an intel machine yet.

Another thing I did try was using PGO, but I wasn't able to see a noticeable difference.

Oh and one more thing I tried was different [bitmap](https://pkg.go.dev/github.com/kelindar/bitmap) indexes for the tags and posts in order to use SIMD operations. However, it's hard to beat the current algorithm for the current data distributions, all my attempts ended up being slower.

### DIFF
--- a/go/main_test.go
+++ b/go/main_test.go
@@ -1,0 +1,15 @@
+package main
+
+import "testing"
+
+// TestMain is useful for tracing a single run:
+// go test -trace go.trace && go tool trace go.trace
+func TestMain(t *testing.T) {
+	main()
+}
+
+func BenchmarkMain(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		main()
+	}
+}
--- a/go_con/main_test.go
+++ b/go_con/main_test.go
@@ -1,0 +1,15 @@
+package main
+
+import "testing"
+
+// TestMain is useful for tracing a single run:
+// go test -trace go.trace && go tool trace go.trace
+func TestMain(t *testing.T) {
+	main()
+}
+
+func BenchmarkMain(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		main()
+	}
+}
--- a/run.sh
+++ b/run.sh
@@ -44,7 +44,7 @@ run_go() {
         cd ./go &&
         go build &&
         if [ $HYPER == 1 ]; then
-            capture "Go" hyperfine -r 10 -w 5 --show-output "./related"
+            capture "Go" hyperfine -r 10 -w 5 --show-output "GOGC=off GOMEMLIMIT=32MiB ./related"
         else
             command ${time} -f '%es %Mk' ./related
         fi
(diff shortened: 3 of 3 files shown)


jinyus:
Hey, thanks for the PR... but total time including I/O has been removed from the table so optimizing it is no longer necessary. It will now show how the program scales when running with 25k and 50k posts. 

Targeting specific hardware is against the [rules](https://github.com/jinyus/related_post_gen#run-benchmark) btw.

felixge:
> Hey, thanks for the PR... but total time including I/O has been removed from the table so optimizing it is no longer necessary. It will now show how the program scales when running with 25k and 50k posts.

Ok, I guess you were moving the goal post as I was working on this 😅 .

> Targeting specific hardware is against the [rules](https://github.com/jinyus/related_post_gen#run-benchmark) btw.

Can you elaborate on what you mean by this? You are running your benchmarks on specific hardware, and compilers/JITs/runtimes/etc. might take advantage of this at both compile time as well as runtime. So what exactly is allowed and not allowed here?

FWIW, I don't think the `GOAMD64` flag is going to make a difference for the inner hot loop in Go. But generally speaking, I'd say taking advantage of the available hardware seems fair game?

felixge:
Also: Any reason why you wouldn't cherry pick 5a665e5c64beecb2743c43ff6626b41fbe99f677 and 76feb8faf54dc7b5409aa1d5d79fd4b3cb8d5a07?

jinyus:
> Ok, I guess you were moving the goal post as I was working on this 😅 .

Yeah sorry about that,

The hardware targeting rule was originally put in place to prevent simd optimizations on x86 cpus. It just goes against the spirit of the competition.

Re the good commits:

Did you see a change when disabling arenas? I will merge the tests

felixge:
> The hardware targeting rule was originally put in place to prevent simd optimizations on x86 cpus. It's just goes against the spirit of the competition.

Ok, thanks for clarifying.

> Re the good commits:
> 
> Did you see a change when disabling arenas? I will merge the tests

arenes were removed 4 days ago in fbe6f2e7380a1a0426dc6957f7578b39b121f2b1. My commit is just cleaning up something that the other commit had forgotten.

PR #171, opened 2023-10-09
Simplify and optimize Julia implementation

LilithHafner:
Two changes:
- Replace `haskey`, `d[k] = x` and `d[k]` with `get!` to cut down on code size and the number of required hashes. `get!` is a builtin function designed specifically for this common use case of getting an element from a dictionary or instantiating it if it's missing.
- Use an appropriately sized integer type for post index and counts. This will cause bugs if there are more than 2^16 posts. If that possibility for error is "cheating", we can add a near-zero-cost dispatch to dynamically determine the appropriate number type. I'd estimate that that dispatch will add about 6 lines of code, increase compilation time a bit, and not affect runtime at all.

### DIFF
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -33,7 +33,7 @@ end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
 
-function fastmaxindex!(xs::Vector{Int64}, topn, maxn, maxv)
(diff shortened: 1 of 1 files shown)


jinyus:
Hey, the [rules](https://github.com/jinyus/related_post_gen#run-benchmark) state that it must be able to handle 100k posts so that tweak is not allowed.

PR #175, opened 2023-10-10
Removed local benchmark (`@btime`) artefact

algunion:
There was a local testing artefact left in the code (`@btime` usage for local benchmark).

Just removed it so this can work well when the benchmark is run.


LilithHafner (approved):
Yikes, thanks. Sorry that that slipped through.

PR #177, opened 2023-10-10
Bring back TreeDictionary

yaglo:
Not sure why this one got deleted. It's an official and stable library: https://github.com/apple/swift-collections, unlike what the rust_con is using with `rust_rayon v0.1.0`, which is now failing with `error[E0658]: use of unstable library feature 'local_key_cell_methods'` ;)

### DIFF
--- a/swift_con/Sources/main.swift
+++ b/swift_con/Sources/main.swift
@@ -1,4 +1,4 @@
-// import struct Collections.TreeDictionary
+import struct Collections.TreeDictionary
 import Foundation
 import Dispatch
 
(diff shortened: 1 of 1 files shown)


jinyus:
I gave the reasons here: https://github.com/jinyus/related_post_gen/pull/95#issuecomment-1747215397

Let me knwo what you think?

yaglo:
Sorry, didn't see your original comment. In terms of the difference, it's about 20-25% on my machine for the non-IO measuments, 7.5ms vs 9.6ms on average.

jinyus:
Alright, update the deps and I'll merge

yaglo:
I don't think they need to be updated, `Package.swift` hasn't changed since the original commit.

PR #178, opened 2023-10-10
Concurrent Julia - testing the waters

algunion:
This is the first attempt at a concurrent Julia version.

### DIFF
--- a/julia_con/related.jl
+++ b/julia_con/related.jl
@@ -1,20 +1,23 @@
+using Base.Threads
+
 using JSON3
 using StructTypes
 using Dates
 using StaticArrays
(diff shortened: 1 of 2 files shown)


algunion (commented):
Thanks. Missed that.

jinyus:
concurrent version is 3x slower for me

algunion:
> concurrent version is 3x slower for me

On my end, the parallel version does 8ms (form 21ms for the non-parallel version).

Is the warming up working properly? 

I can push a version with the warming call (so we make sure we do not count the compiler time).

jinyus:
I added the warm up call. It's 7ms now

PR #182, opened 2023-10-10
[Julia] move `topn` definition into body of `@threads`

LilithHafner:
Julia is worried that `topn` might be mutated and so deoptimizes. The Julia people should really figure this out because there is literally only one assignment to `topn`, but whatever, moving the definition into the body of the loop speeds things up.

I also removed trailing whitespace.

### DIFF
--- a/julia_con/related.jl
+++ b/julia_con/related.jl
@@ -69,7 +69,6 @@ function related(posts)
     end
 end
 function related(::Type{T}, posts) where {T}
-    topn = 5
     # key is every possible "tag" used in all posts
     # value is indicies of all "post"s that used this tag
     tagmap = Dict{String,Vector{T}}()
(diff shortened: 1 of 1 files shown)


algunion:
> Julia is worried that `topn` might be mutated and so deoptimizes. The Julia people should really figure this out because there is literally only one assignment to `topn`, but whatever, moving the definition into the body of the loop speeds things up.

This _small_ (well - not so small given the deoptimization effect) thing speeds things up by about 20% (8ms -> a little over 6ms on my machine).

Tortar:
Is this the same https://discourse.julialang.org/t/using-an-assignment-outside-of-threaded-for-loop-degraded-performance-by-a-huge-amount/105552 but much worse?

PR #191, opened 2023-10-11
Revert "formatting Julia and remove unnecessary deps"

Moelf:
it's possible my weird tweaking introduced performance degradation.

I don't fully understand Github CI environment, so let's revert it to be safe

### DIFF
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -1,7 +1,13 @@
-using JSON3, Dates, StaticArrays
+using JSON3
+using StructTypes
+using Dates
+using StaticArrays
+
+# warmup is done by hyperfine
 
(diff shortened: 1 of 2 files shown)


Moelf:
it's still some vCPU shenanigan so idk, we can revert and see how it goes maybe

Moelf:
so weird ... @algunion

algunion:
> so weird ... @algunion

I have no idea what is going on :)

It seems that on my end I need to learn how to go deeper in the profiling side and not rely only on the benchmark numbers. Especially when there are small differences and the runs have some variation, it is hard to say if a change did some deoptimization or not.

PR #193, opened 2023-10-11
Added D regular implementation

cyrusmsk:
Added D implementation based on the code from Go.
I've tested several D implementations. This one based on memutils package for HashMap, which worked a bit better than built-in Associative Arrays.

Parallel versions also under preparation, but currently they are working quite slow.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM archlinux:base
 # Update package repository
 RUN pacman -Syu --noconfirm
 
-RUN pacman -S --noconfirm --needed wget unzip sudo base-devel git go clang llvm python python-pip ncurses gcc llvm hyperfine rustup dotnet-sdk crystal zig julia dart nodejs deno maven nim opam dune lua51 luajit luarocks libedit github-cli less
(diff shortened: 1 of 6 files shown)


cyrusmsk:
> I don't know much about D but wouldn't it need to install the deps before compiling?

when you run "dub build" it automatically install deps into specific folder and use it flawlessly.
I've updated Dockerfile to add 'dub' (D package manager). Because on Archlinux it is not in the bundle of compiler package.

jinyus:
I also updated the build command to 

```
dub build --build=release
```

as that's what works for me. I notice D has an official compiler (DMD). Is it that ldc is faster?

cyrusmsk:
> I also updated the build command to
> 
> ```
> dub build --build=release
> ```
> 
> as that's what worked for me. I notice D has an official compiler (DMD). Is it that ldc is faster?

DMD is a reference compiler with its own backend. LDC is using DMD's frontend and LLVM backend. There is also GDC - which uses GCC as a backend. LDC/GDC could provide more optimizations in the code. So it is better to use one of them. It is possible even prepare 2 runs for both of them.

PR #195, opened 2023-10-11
[csharp] Changed Clear() to Fill(0)

mark-plunkett:
Around 9% faster on my machine

```
dotnet publish -c Release
1..1000 | %{ measure-command { .\bin\release\net7.0\win-x64\native\related.exe } | % TotalMilliseconds } | measure -Average
```

- With `Fill(0)`: 79.4570753ms
- With `Clear()`: 86.0791141ms


mark-plunkett:
Looks like this made things slower on the VM benchmarks though

PR #198, opened 2023-10-12
Separate JIT and AOT configs for C#

MichalStrehovsky:
While AOT finishes faster for the 5k entries JSON in the repo, it is not faster for larger sets, at least in .NET 7. Benchmark both flavors like for Java or Dart.

Things look different in .NET 8 (I see .NET 8 AOT matching Go speed) but the repo didn't update to .NET 8 yet. It could - .NET 8 RC2 is supported for use in production by Microsoft (marked [go-live](https://dotnet.microsoft.com/en-us/download/dotnet/8.0)).

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -89,6 +89,9 @@ jobs:
 
             - name: Install Dotnet
               uses: actions/setup-dotnet@v3
+              with:
+                dotnet-version: '8.0.x'
+                dotnet-quality: 'preview'
 
             - name: Install LuaJit
               run: |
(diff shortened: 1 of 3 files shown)


jinyus:
Ok, you can upgrade it to .net8

MichalStrehovsky:
> Ok, you can upgrade it to .net8

Bumped to 8.0 as well. I _think_ the Github install action still needs to specify `dotnet-quality` as preview, but this should install RC2.

jinyus:
That's fine, I no longer run the benchmarks on github actions.

PR #200, opened 2023-10-12
[Julia] Remove StaticArrays.jl, replace with Vector5 types

### LINKED ISSUE #194: Julia's StaticArrays contains unsafe code
Hi,

it seems that Julia solution uses `StaticArrays.jl` package which uses unsafe code https://github.com/JuliaArrays/StaticArrays.jl/blob/d80455d2e1aba57a6cf65cc8d787dc6474ec3980/src/MArray.jl#L25

Honestly my understanding of the source of the package is rather poor
so maybe the unsafe part is not called in the solution.

But if the unsafe part is called then it probably violates the rules.

mkitti:
This entirely removes StaticArrays.jl from the Julia code. Instead, we implement new {M,S}Vector5 types specialized for mutable or immutable 5-element collections.

Also, this replaces manual package management with Project.toml files including the package UUIDs, adding specificity and additional security.

Fixes #194

### DIFF
--- a/julia_con/related.jl
+++ b/julia_con/related.jl
@@ -1,6 +1,3 @@
-using Pkg
-Pkg.instantiate()
-
 using Base.Threads
 
 using JSON3
--- a/julia_con/related.jl
+++ b/julia_con/related.jl
@@ -84,8 +84,8 @@ function related(::Type{T}, posts) where {T}
 
     @threads for (postsrange, _) in chunks(posts, nthreads())
         topn = 5
-        maxn = Vector{T}(undef,topn)
-        maxv = Vector{T}(undef,topn)
+        maxn = MVector5{T}()
+        maxv = MVector5{T}()
         taggedpostcount = Vector{T}(undef, length(posts))
 
         for i in postsrange
--- a/julia_con/related.jl
+++ b/julia_con/related.jl
@@ -1,12 +1,16 @@
+using Pkg
+Pkg.instantiate()
+
 using Base.Threads
 
 using JSON3
 using StructTypes
 using Dates
-using StaticArrays
 using ChunkSplitters
 
 # warmup is done by hyperfine
+
+include("../julia/vector5.jl")
 
 function relatedIO()
     json_string = read("../posts.json", String)
@@ -33,7 +37,7 @@ end
 struct RelatedPost
     _id::String
     tags::Vector{String}
-    related::SVector{5,PostData}
+    related::SVector5{PostData}
 end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
@@ -105,7 +109,7 @@ function related(::Type{T}, posts) where {T}
 
             fastmaxindex!(taggedpostcount, topn, maxn, maxv)
 
-            relatedpost = RelatedPost(post._id, post.tags, SVector{topn}(@view posts[maxn]))
+            relatedpost = RelatedPost(post._id, post.tags, SVector5(@view posts[maxn]))
             relatedposts[i] = relatedpost
         end
     end
--- a/julia/Project.toml
+++ b/julia/Project.toml
@@ -1,0 +1,5 @@
+[deps]
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+StructTypes = "856f2bd8-1eba-4b0a-8007-ebc267875bd4"
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -1,9 +1,12 @@
+using Pkg
+Pkg.instantiate()
 using JSON3
 using StructTypes
 using Dates
-using StaticArrays
 
 # warmup is done by hyperfine
+
+include("vector5.jl")
 
 function relatedIO()
     json_string = read("../posts.json", String)
@@ -30,7 +33,7 @@ end
 struct RelatedPost
     _id::String
     tags::Vector{String}
-    related::SVector{5,PostData}
+    related::SVector5{PostData}
 end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
@@ -80,8 +83,8 @@ function related(::Type{T}, posts) where {T}
     relatedposts = Vector{RelatedPost}(undef, length(posts))
     taggedpostcount = Vector{T}(undef, length(posts))
 
-    maxn = Vector{T}(undef,topn)
-    maxv = Vector{T}(undef,topn)
+    maxn = MVector5{T}()
+    maxv = MVector5{T}()
 
     for (i, post) in enumerate(posts)
         taggedpostcount .= 0
@@ -99,7 +102,7 @@ function related(::Type{T}, posts) where {T}
 
         fastmaxindex!(taggedpostcount, topn, maxn, maxv)
 
-        relatedpost = RelatedPost(post._id, post.tags, SVector{topn}(@view posts[maxn]))
+        relatedpost = RelatedPost(post._id, post.tags, SVector5(@view posts[maxn]))
         relatedposts[i] = relatedpost
     end
 
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -1,5 +1,3 @@
-using Pkg
-Pkg.instantiate()
 using JSON3
 using StructTypes
 using Dates
--- a/julia/vector5.jl
+++ b/julia/vector5.jl
@@ -1,0 +1,38 @@
+abstract type Vector5{T} <: AbstractVector{T} end
+Base.size(::Vector5) = (5,)
+Base.IndexStyle(::Type{<: Vector5}) = IndexLinear()
+Base.getindex(A::Vector5, i::Int) = 
+    i == 1 ? A.a :
+    i == 2 ? A.b :
+    i == 3 ? A.c :
+    i == 4 ? A.d :
+    i == 5 ? A.e :
+    throw(BoundsError(A, i))
+
+mutable struct MVector5{T} <: Vector5{T}
+    a::T
+    b::T
+    c::T
+    d::T
+    e::T
+end
+MVector5{T}() where T = MVector5{T}(zero(T), zero(T), zero(T), zero(T), zero(T))
+Base.setindex!(A::MVector5, v, i::Int) =
+    i == 1 ? A.a = v :
+    i == 2 ? A.b = v :
+    i == 3 ? A.c = v :
+    i == 4 ? A.d = v :
+    i == 5 ? A.e = v :
+    throw(BoundsError(A, i))
+
+struct SVector5{T} <: Vector5{T}
+    a::T
+    b::T
+    c::T
+    d::T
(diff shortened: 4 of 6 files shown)


julia_con/related.jl
```diff
@@ -1,13 +1,17 @@
...
 
 using JSON3
 using StructTypes
 using Dates
-using StaticArrays
 using ChunkSplitters
 
 # warmup is done by hyperfine
 
+include("../julia/vector5.jl")
```
mkitti, line 13: Is this permissible or should I duplicate the file?

jdiaz97:
Looks good

jinyus:
I not sure how I feel about this. Making a specialized, fixed size array definitely goes against the spirit of the competition and other solutions will mimic this to level the playing field.

With that said, I'm not seeing a speed up on my end. This and the old solution is pretty much equal(linux). How are the results looking for you?

old:

```
Benchmark 1: julia related.jl
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 21 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
  Time (mean ± σ):      2.011 s ±  0.019 s    [User: 1.860 s, System: 0.143 s]
  Range (min … max):    1.966 s …  2.033 s    10 runs
```  

new:

```
Benchmark 1: julia --project related.jl
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
Processing time (w/o IO): 23 milliseconds
  Time (mean ± σ):      1.728 s ±  0.024 s    [User: 1.569 s, System: 0.150 s]
  Range (min … max):    1.700 s …  1.767 s    10 runs
```

mkitti:
I'm running on a potato at the moment, but I do see an impact versus main.

Old:
```
Benchmark 1: julia related.jl
Processing time (w/o IO): 57 milliseconds
Processing time (w/o IO): 68 milliseconds
Processing time (w/o IO): 57 milliseconds
Processing time (w/o IO): 68 milliseconds
Processing time (w/o IO): 55 milliseconds
Processing time (w/o IO): 57 milliseconds
Processing time (w/o IO): 67 milliseconds
Processing time (w/o IO): 68 milliseconds
Processing time (w/o IO): 68 milliseconds
Processing time (w/o IO): 57 milliseconds
Processing time (w/o IO): 68 milliseconds
Processing time (w/o IO): 69 milliseconds
Processing time (w/o IO): 56 milliseconds
  Time (mean ± σ):      4.232 s ±  0.018 s    [User: 4.037 s, System: 0.380 s]
  Range (min … max):    4.204 s …  4.262 s    10 runs
```

New:
```
Benchmark 1: julia --project related.jl
Processing time (w/o IO): 49 milliseconds
Processing time (w/o IO): 48 milliseconds
Processing time (w/o IO): 48 milliseconds
Processing time (w/o IO): 49 milliseconds
Processing time (w/o IO): 49 milliseconds
Processing time (w/o IO): 49 milliseconds
Processing time (w/o IO): 48 milliseconds
Processing time (w/o IO): 48 milliseconds
Processing time (w/o IO): 50 milliseconds
Processing time (w/o IO): 49 milliseconds
Processing time (w/o IO): 50 milliseconds
Processing time (w/o IO): 49 milliseconds
Processing time (w/o IO): 49 milliseconds
  Time (mean ± σ):      3.634 s ±  0.078 s    [User: 3.480 s, System: 0.339 s]
  Range (min … max):    3.551 s …  3.795 s    10 runs
```

jinyus:
Just tested it on the benchmarking VM. Slower at 5k, same at 20k and a tap slower at 60k:

old:

```
5k -> 34ms
20k -> 576ms
60k -> 5.04s

```

new:

```
Julia:

        Benchmark 1: julia --project related.jl
        Processing time (w/o IO): 39 milliseconds
        Processing time (w/o IO): 39 milliseconds
        Processing time (w/o IO): 38 milliseconds
        Processing time (w/o IO): 39 milliseconds
        Processing time (w/o IO): 39 milliseconds
        Processing time (w/o IO): 38 milliseconds
        Processing time (w/o IO): 39 milliseconds
        Processing time (w/o IO): 38 milliseconds
        Processing time (w/o IO): 38 milliseconds
        Processing time (w/o IO): 38 milliseconds
        Processing time (w/o IO): 38 milliseconds
        Processing time (w/o IO): 39 milliseconds
        Processing time (w/o IO): 38 milliseconds
          Time (mean ± σ):      2.800 s ±  0.030 s    [User: 2.622 s, System: 0.177 s]
          Range (min … max):    2.771 s …  2.861 s    10 runs
         
Julia:

        Benchmark 1: julia --project related.jl
        Processing time (w/o IO): 577 milliseconds
        Processing time (w/o IO): 577 milliseconds
        Processing time (w/o IO): 577 milliseconds
          Time (mean ± σ):      4.009 s ±  0.004 s    [User: 3.822 s, System: 0.186 s]
          Range (min … max):    4.007 s …  4.012 s    2 runs
         
Julia:

        Benchmark 1: julia --project related.jl
        Processing time (w/o IO): 5077 milliseconds
        Processing time (w/o IO): 5066 milliseconds
        Processing time (w/o IO): 5058 milliseconds
          Time (mean ± σ):     13.344 s ±  0.003 s    [User: 13.052 s, System: 0.282 s]
          Range (min … max):   13.342 s … 13.346 s    2 runs

```

julia_con/related.jl
jinyus: all good

mkitti:
Hmm... I guess not much point in terms of performance then. Let's hold on for the moment. I see some things to try.

MasonProtter:
Why not just use a `NTuple`?

mkitti:
For the SVector5 implementation, sure, but we also need a `MVector` replacement.

MasonProtter (commented):
you can just invoke `getfield` and `setfield!` instead of branching.

MasonProtter:
Hm, nevermind, after doing some benchmarks on my machine, the above suggestion I posted is actually slower. I also observe @mkitti's implementation here recovering most of the performance lost when StaticArrays was dropped.

mkitti:
I had tried a version of this where I inline the call to get/setfield and things seem to get worse.

The problem is that the field functions have to do another check to validate the input and possibly throw. I think our branching ends up being cheaper.

I have also prepared a package refactoring to take advantage of precompile disk caching in the form of pkgimages:

https://github.com/mkitti/related_post_gen/pull/1

algunion:
> I not sure how I feel about this. Making a specialized, fixed size array definitely goes against the spirit of the competition and other solutions will mimic this to level the playing field.

@jinyus,

Consider this: we already have this in the Julia ecosystem (e.g., `StaticArrays.jl`). The package is blazing fast and safe (take a look at [arguments ](https://github.com/jinyus/related_post_gen/issues/194#issuecomment-1759217426)like this one).

Now, you are saying that if @mkitti implements his code in this repo, this feels like going against the spirit of the competition. But if we go and publish it on the Julia registry (and it becomes something more general like having a Julia macro that generates specialized code for any small N - 5 included - and we call it as a package, then we are OK).

Also, if people are to invest more time in this benchmark, I think we need to have some transparent and predictable rules and also make sure those rules are consistently applied to all languages involved. See [here](https://github.com/jinyus/related_post_gen/issues/203).

Now it seems a new rule appeared: if you do not have a package that does X, you are also not allowed to implement X - even if you do it in a transparent and clearly safe way.

MasonProtter:
Maybe superseded by https://github.com/jinyus/related_post_gen/pull/207 which is a more general approach and doesn't require hand-made structs of a specific size.

dlakelan:
The reason for this is that StaticArrays.jl was ruled as breaking the contests rules because it "invokes unsafe code". So this is just a reimplementation of the subset of functionality needed from StaticArrays but without invoking any "unsafe" code...

for those who haven't been following the relevant discussion is here: 

https://github.com/jinyus/related_post_gen/issues/194

jinyus:
@algunion 

His approach  abides by the rules but it goes against the spirit of the competition because it's trying to win by avoiding work that every other language is doing. If his approach was faster, it would be implemented in all the other languages to level the playing field.

mkitti:
All that we're trying to do is the equivalent of the following C code. That is we want to avoid allocating an array on the heap by allocating on the stack.

```C
#include<stdint.h>

int main() {
    int topn = 5;
    uint32_t maxn[topn];
    uint32_t maxv[topn];

    // ...

    return 0;
}
```

You might argue that if such statically sized arrays were a feature of Julia that it should be part of the standard library. However, Julia development has a particular [philosophy](https://discourse.julialang.org/t/when-should-a-package-move-to-the-standard-library-or-system-image-staticarrays-what-is-it/73577/29?u=mkitti) of [minimalism](https://github.com/JuliaLang/julia/issues/18795). That is if an independent package can implement the feature, then the feature should not be implemented by the standard library. In fact there is a concerted [excision](https://github.com/JuliaLang/julia/labels/excision) effort to make the `Base` package smaller and turn standard library packages into just normal packages that happen to be installed by default. A strong driver for this is that independent packages can iterate  faster than the language and standard library can.

https://docs.julialang.org/en/v1/manual/performance-tips/#Consider-StaticArrays.jl-for-small-fixed-size-vector/matrix-operations

As I mentioned above, StaticArrays.jl is also [recommended in the primary manual](https://docs.julialang.org/en/v1/manual/performance-tips/#Consider-StaticArrays.jl-for-small-fixed-size-vector/matrix-operations) for Julia as a performance tip. There indeed have been calls in the past to turn [StaticArrays.jl into a standard library package](https://discourse.julialang.org/t/when-should-a-package-move-to-the-standard-library-or-system-image-staticarrays-what-is-it/73577/1). This is why some of my compatriots are quite frustrated because the rules here prevent the use of what has become a standard optimization step.

> His approach abides by the rules but it goes against the spirit of the competition because it's trying to win by avoiding work that every other language is doing.

I'm unclear what work my approach avoids that every other language is using. We are just trying to implement a static fixed sized array where the size of the array is known at compile time. In some languages such as C, it is standard feature. In such a case, the compiler can often perform optimizations with that information. _Our goal here is to access those same optimizations that are available to other languages._

What I have implemented here is a [minimal abstract array interface](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array) on top of a mutable stack allocated type to see if doing so alone might be lead to some performance gains. It is a slightly distinct version from the `MVector` implementation in StaticArrays.jl in that each element is also addressable and mutable as a named field, which means that I do not need to do the potentially "unsafe" pointer operations.

To summarize, we are not trying to do something that the other languages cannot or are not doing. Rather, we want to take advantage of the same optimizations that the other languages are using by allocating fixed size arrays on the stack. In Julia, allocated fixed arrays on the stack is an operation that can be and thus has been delegated to an independent package. Because of your concerns with the implementation of that package, I am now trying to do a minimal implementation of the same concept without the concerning "unsafe" functions.

jinyus:
I'll put this into drafts for now. Let me know when you're ready to merge.

PR #204, opened 2023-10-12
Some cleanup and a more generalized approach with Insertion sort

DrBlury:
About 5-20% faster, cleaner code and more generalized approach

### DIFF
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -9,6 +9,7 @@ function relatedIO()
     json_string = read("../posts.json", String)
     posts = JSON3.read(json_string, Vector{PostData})    
 
+    related(posts)
     start = now()
     all_related_posts = related(posts)
     println("Processing time (w/o IO): $(now() - start)")
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -9,9 +9,8 @@ function relatedIO()
     json_string = read("../posts.json", String)
     posts = JSON3.read(json_string, Vector{PostData})    
 
-    related(posts)
-
     start = now()
+    related(posts)
     all_related_posts = related(posts)
     println("Processing time (w/o IO): $(now() - start)")
 
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -10,7 +10,6 @@ function relatedIO()
     posts = JSON3.read(json_string, Vector{PostData})    
 
     start = now()
-    related(posts)
     all_related_posts = related(posts)
     println("Processing time (w/o IO): $(now() - start)")
 
--- a/go/go.sum
+++ b/go/go.sum
@@ -1,4 +1,0 @@
-github.com/goccy/go-json v0.10.2 h1:CrxCmQqYDkv1z7lO7Wbh2HN93uovUHgrECaO5ZrCXAU=
-github.com/goccy/go-json v0.10.2/go.mod h1:6MelG93GURQebXPDq3khkgXZkazVtN9CRI+MGFi0w8I=
-github.com/ugurcsen/gods-generic v0.10.4 h1:OomH3R2MdzZxpnEPijaD/ncLzV6rpDXd5ruEkWsw0vo=
-github.com/ugurcsen/gods-generic v0.10.4/go.mod h1:mGYOa88Y5sbw+ADXLpScxjJ7s5iHoWya/YHyeQ4f6c4=
--- a/go/main.go
+++ b/go/main.go
@@ -45,9 +45,11 @@ func computeRelatedPosts(posts []Post, tagMap map[string][]isize) []RelatedPosts
 	taggedPostCount := make([]isize, postsLen)
 
 	for i := range posts {
+		// Reset the tag count for each post
 		for j := range taggedPostCount {
 			taggedPostCount[j] = 0
 		}
+
 		// Count the number of tags shared between posts
 		for _, tag := range posts[i].Tags {
 			for _, otherPostIdx := range tagMap[tag] {
@@ -55,28 +57,22 @@ func computeRelatedPosts(posts []Post, tagMap map[string][]isize) []RelatedPosts
 			}
 		}
 		taggedPostCount[i] = 0 // Don't count self
+
 		top5 := [topN]PostWithSharedTags{}
-		minTags := isize(0)
 
 		for j, count := range taggedPostCount {
-			if count > minTags {
-				// Find the position to insert
-				pos := 4
-				for pos >= 0 && top5[pos].SharedTags < count {
-					pos--
-				}
-				pos++
+			if count > top5[0].SharedTags {
+				top5[0] = PostWithSharedTags{Post: isize(j), SharedTags: count}
 
-				// Shift and insert
-				if pos < 4 {
-					copy(top5[pos+1:], top5[pos:4])
+				// Insertion sort to maintain the order
+				for k := 0; k < topN-1; k++ {
+					if top5[k].SharedTags > top5[k+1].SharedTags {
+						top5[k], top5[k+1] = top5[k+1], top5[k]
+					}
 				}
-
-				top5[pos] = PostWithSharedTags{Post: isize(j), SharedTags: count}
-				minTags = top5[4].SharedTags
 			}
 		}
-		// Convert indexes back to Post pointers
+
 		topPosts := [topN]*Post{}
 		for idx, t := range top5 {
 			topPosts[idx] = &posts[t.Post]
@@ -88,6 +84,7 @@ func computeRelatedPosts(posts []Post, tagMap map[string][]isize) []RelatedPosts
 			Related: topPosts,
 		}
 	}
+
 	return allRelatedPosts
 }
 
--- a/go/main.go
+++ b/go/main.go
@@ -40,9 +40,8 @@ func buildTagMap(posts []Post) map[string][]isize {
 }
 
 func computeRelatedPosts(posts []Post, tagMap map[string][]isize) []RelatedPosts {
-	postsLen := len(posts)
-	allRelatedPosts := make([]RelatedPosts, postsLen)
-	taggedPostCount := make([]isize, postsLen)
+	allRelatedPosts := make([]RelatedPosts, 0, len(posts))
+	taggedPostCount := make([]isize, len(posts))
 
 	for i := range posts {
 		// Reset the tag count for each post
@@ -78,11 +77,11 @@ func computeRelatedPosts(posts []Post, tagMap map[string][]isize) []RelatedPosts
 			topPosts[idx] = &posts[t.Post]
 		}
 
-		allRelatedPosts[i] = RelatedPosts{
+		allRelatedPosts = append(allRelatedPosts, RelatedPosts{
 			ID:      posts[i].ID,
 			Tags:    &posts[i].Tags,
 			Related: topPosts,
-		}
+		})
 	}
 
 	return allRelatedPosts
--- a/go/main.go
+++ b/go/main.go
@@ -7,12 +7,11 @@ import (
 	"time"
 )
 
-const topN = 5
-const InitialTagMapSize = 100
-
 const (
-	INPUT_FILE  = "../posts.json"
-	OUTPUT_FILE = "../related_posts_go.json"
+	topN              = 5
+	InitialTagMapSize = 100
+	INPUT_FILE        = "../posts.json"
+	OUTPUT_FILE       = "../related_posts_go.json"
 )
 
 type isize uint16
--- a/go/main.go
+++ b/go/main.go
@@ -9,9 +9,8 @@ import (
 
 const topN = 5
 const InitialTagMapSize = 100
-const InitialPostsSliceCap = 0
 
-type isize uint32
+type isize uint16
 
 type Post struct {
 	ID    string   `json:"_id"`
@@ -30,24 +29,18 @@ type PostWithSharedTags struct {
 	SharedTags isize
 }
 
-func main() {
-	file, _ := os.Open("../posts.json")
-	var posts = make([]Post, 0, InitialPostsSliceCap)
-	err := json.NewDecoder(file).Decode(&posts)
-	if err != nil {
-		fmt.Println(err)
-	}
-	postsLen := len(posts)
-	start := time.Now()
-	// assumes that there are less than 100 tags
+func buildTagMap(posts []Post) map[string][]isize {
 	tagMap := make(map[string][]isize, InitialTagMapSize)
-
 	for i, post := range posts {
 		for _, tag := range post.Tags {
 			tagMap[tag] = append(tagMap[tag], isize(i))
 		}
 	}
+	return tagMap
+}
 
+func computeRelatedPosts(posts []Post, tagMap map[string][]isize) []RelatedPosts {
+	postsLen := len(posts)
 	allRelatedPosts := make([]RelatedPosts, postsLen)
 	taggedPostCount := make([]isize, postsLen)
 
@@ -63,7 +56,7 @@ func main() {
 		}
 		taggedPostCount[i] = 0 // Don't count self
 		top5 := [topN]PostWithSharedTags{}
-		minTags := isize(0) // Updated initialization
+		minTags := isize(0)
 
 		for j, count := range taggedPostCount {
 			if count > minTags {
@@ -95,8 +88,27 @@ func main() {
 			Related: topPosts,
 		}
 	}
+	return allRelatedPosts
+}
+
+func readPosts() []Post {
+	file, _ := os.Open("../posts.json")
+	var posts = []Post{}
+	err := json.NewDecoder(file).Decode(&posts)
+	if err != nil {
+		fmt.Println(err)
+		return nil
+	}
+	return posts
+}
+
+func main() {
+	posts := readPosts()
+	start := time.Now()
+	tagMap := buildTagMap(posts)
+	allRelatedPosts := computeRelatedPosts(posts, tagMap)
 
 	fmt.Println("Processing time (w/o IO):", time.Since(start))
-	file, _ = os.Create("../related_posts_go.json")
+	file, _ := os.Create("../related_posts_go.json")
 	_ = json.NewEncoder(file).Encode(allRelatedPosts)
 }
--- a/go/main.go
+++ b/go/main.go
@@ -10,12 +10,17 @@ import (
 const topN = 5
 const InitialTagMapSize = 100
 
+const (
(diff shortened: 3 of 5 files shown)


DrBlury:
Notice I moved the julia type assertion into the compute time because it's not i/o.

julia/related.jl
```diff
@@ -9,9 +9,8 @@ function relatedIO()
     json_string = read("../posts.json", String)
     posts = JSON3.read(json_string, Vector{PostData})    
 
-    related(posts)
```
algunion, line 12: Why remove the compilation call and include it in the timing?

algunion (changes requested):

algunion:
> Notice I moved the julia type assertion into the compute time because it's not i/o.

No, you didn't :)

The Julia time will be now: 2 complete calls of the `related` function + the compilation time of the `related` function (and the other functions called by the `related` function).

I am convinced that you are well intended... but I think it would be fair to do minimal research before deciding to make that kind of change.

julia/related.jl
DrBlury: As the size of the posts is only known during processing (after i/o) any type optimizations have to be measured as well. 
```

function related(posts)
    for T in (UInt8, UInt16, UInt32, UInt64)
        if length(posts) < typemax(T)
            return related(T, posts)
        end
    end
end

```
Is not done during compilation as I understand. It just asserts the type (UInt8, UInt32 etc.) from the posts size.

Moelf:
if you're unhappy with Julia's manual warmup, we can revive:
- https://github.com/jinyus/related_post_gen/pull/163


Basically to have native code cached, you want to put the thing into a project. We wanted to avoiding making more directory structures so we opted for a manual call as warmup and cache native code. It does not cache result or value, just a way to trigger JIT compilation.

julia/related.jl
algunion: So this "as I understand" of yours somehow translates to the solution of "let's have the full compilation code + 2x runs added to this language result"?

I am trying to be respectful here: this conversation we have here should not occur. And that change you made should never have been done.

Moelf: you're still removing the manual warmup.

This function you're looking at is always called at runtime.

MasonProtter: What could be done instead is do 4 calls before the timing loop, e.g.
```
related(UInt8, posts[1:20])
related(UInt16, posts[1:20])
related(UInt32, posts[1:20])
related(UInt64, posts[1:20])
```
then every relevant code path should be hit and compiled on a small workload before the timing loop begins.

julia/related.jl
```diff
@@ -10,7 +10,6 @@ function relatedIO()
     posts = JSON3.read(json_string, Vector{PostData})    
 
     start = now()
-    related(posts)
```
algunion, line 12: If you remove the warmup call then we are arriving to something similar to this:

image

This is properly [documented ](https://docs.julialang.org/en/v1/manual/performance-tips/#Measure-performance-with-[@time](@ref)-and-pay-attention-to-memory-allocation) here.

If you are stumbling upon the runtime dispatch of the unsigned int types, that has nothing to do with this.

Please keep in mind that Julia does not compile the code in the same manner as Go/Rust.

You can read more about this [here](https://docs.julialang.org/en/v1/manual/performance-tips/#Measure-performance-with-[@time](@ref)-and-pay-attention-to-memory-allocation).

DrBlury: Alright, I understood that this is not the type assertions that this is used for but for warming up (compiling the function using the JIT compiler for later use). But are we sure that there is no caching happening? The type assertions for example have to be run during the measured time. But if the function is warmed up with real data, it may already cache some data.

Also: I started a discussion issue:
https://github.com/jinyus/related_post_gen/issues/205

Moelf: >But are we sure that there is no caching happening? 

the only thing get's cached is the machine code of the function. When you call it the next time, it still needs to check length of `posts` and do a runtime dynamic dispatch

MasonProtter: Julia's JIT is not like most traditional hotspot jits. It doesn't cache based on runtime values (unless one explicitly lifts those values to the type domain).

DrBlury: So would it be fine to do the warmup without the actual data like this:
```
# warum up without the actual data
related(PostData[])
```

instead of 
`related(posts)`

Moelf: yes, it would absolutely be fine

algunion: > yes, it would absolutely be fine

I think we need to test it. I am talking about some conditional function calls that can be impacted by the size of the data.

Just a silly example:
```julia
f() = "something that needs compiling"

function f1(xs)
    for (i, _) in enumerate(xs)
        if i > 25 # this number can be runtime dependend
            f()
        end
    end
end
```

If we call `f1` with something with a size less than 26, the `f` function might not be called (and compiled) at all.

Not saying this is actually the case with our current scenario - just saying that such "small" calls might work or might not.

Moelf: the single-argument `related()` is just a run-time dispatcher, it ends up calling one of:
```julia
related(UInt8, posts)
related(UInt16, posts)
related(UInt32, posts)
related(UInt64, posts)
```

they do the real work, if we force compile each of them, I don't see it going wrong

Moelf: you may also have a slightly off understanding of Julia compile model:
```julia
julia> @generated function f(x)
           println("this is compile time for $x")
           return 1
       end
f (generic function with 1 method)

julia> function f1(xs)
           for (i, _) in enumerate(xs)
               if i > 25 # this number can be runtime dependend
                   f(1)
               end
           end
       end
f1 (generic function with 1 method)

julia> f1([1])
this is compile time for Int64
```

DrBlury: I checked it right now and it seems to be equivalent in speed to me

mikmoore:
```
for T in (UInt8, UInt16, UInt32, UInt64)
````
I assume someone has benchmarked this and verified that there truly is a performance improvement for narrower types? It seems that there isn't a lot of room for SIMD here so I wouldn't have expected much difference.

Comment (DrBlury) [reply julia/related.jl:12]: SIMD is not allowed: https://github.com/jinyus/related_post_gen/issues/126
Also:
https://github.com/jinyus/related_post_gen/pull/166#issuecomment-1752179043

**Comment by DrBlury** [comment]
That is very odd. On my machine:
Old:
Benchmark 1: ./related
```
Processing time (w/o IO): 23.913833ms
Processing time (w/o IO): 18.819125ms
Processing time (w/o IO): 18.387583ms
Processing time (w/o IO): 18.44075ms
Processing time (w/o IO): 18.353083ms
Processing time (w/o IO): 19.247333ms
Processing time (w/o IO): 18.802625ms
Processing time (w/o IO): 19.169666ms
Processing time (w/o IO): 18.553125ms
Processing time (w/o IO): 18.385917ms
Processing time (w/o IO): 18.409042ms
Processing time (w/o IO): 18.298292ms
Processing time (w/o IO): 18.476041ms
Processing time (w/o IO): 18.835792ms
Processing time (w/o IO): 18.533083ms
  Time (mean ± σ):      33.5 ms ±   0.6 ms    [User: 31.7 ms, System: 4.5 ms]
  Range (min … max):    32.6 ms …  34.5 ms    10 runs
```

New:
```
Processing time (w/o IO): 19.9455ms
Processing time (w/o IO): 17.655834ms
Processing time (w/o IO): 17.452ms
Processing time (w/o IO): 17.527083ms
Processing time (w/o IO): 17.534792ms
Processing time (w/o IO): 17.525375ms
Processing time (w/o IO): 17.648292ms
Processing time (w/o IO): 17.501708ms
Processing time (w/o IO): 17.549584ms
Processing time (w/o IO): 17.501291ms
Processing time (w/o IO): 17.597291ms
Processing time (w/o IO): 17.441834ms
Processing time (w/o IO): 18.216708ms
  Time (mean ± σ):      39.2 ms ±   0.8 ms    [User: 38.2 ms, System: 3.6 ms]
  Range (min … max):    38.5 ms …  41.3 ms    10 runs
```

Which is about 5% faster, a few colleagues reported up to 18% increases

**Comment by jinyus** [comment]
I'll run on the benchmark VM and report back.

**Comment by jinyus** [comment]
Results from the VM (`go version go1.21.3 linux/amd64`):

old: `docker run -e TEST_NAME='go' -it --rm databench`

```
Go:

        Benchmark 1: ./related
        Processing time (w/o IO): 24.101841ms
        Processing time (w/o IO): 24.390945ms
        Processing time (w/o IO): 24.185442ms
        Processing time (w/o IO): 24.168141ms
        Processing time (w/o IO): 24.10044ms
        Processing time (w/o IO): 24.05904ms
        Processing time (w/o IO): 24.938753ms
        Processing time (w/o IO): 24.189342ms
        Processing time (w/o IO): 25.076554ms
        Processing time (w/o IO): 24.07584ms
        Processing time (w/o IO): 24.07574ms
        Processing time (w/o IO): 24.290144ms
        Processing time (w/o IO): 24.014639ms
          Time (mean ± σ):      63.0 ms ±   1.1 ms    [User: 58.9 ms, System: 11.0 ms]
          Range (min … max):    61.3 ms …  64.7 ms    10 runs
         
Go:

        Benchmark 1: ./related
        Processing time (w/o IO): 371.413354ms
        Processing time (w/o IO): 370.940647ms
        Processing time (w/o IO): 372.894674ms
          Time (mean ± σ):     518.4 ms ±   8.6 ms    [User: 515.8 ms, System: 33.9 ms]
          Range (min … max):   512.3 ms … 524.5 ms    2 runs
         
Go:

        Benchmark 1: ./related
        Processing time (w/o IO): 3.276354916s
        Processing time (w/o IO): 3.286848069s
        Processing time (w/o IO): 3.276418201s
          Time (mean ± σ):      3.765 s ±  0.003 s    [User: 3.787 s, System: 0.118 s]
          Range (min … max):    3.762 s …  3.767 s    2 runs
```

new `docker run -e GIT_REPO="https://github.com/DrBlury/related_post_gen.git" -e TEST_NAME='go' -it --rm databench`:

```
Go:

        Benchmark 1: ./related
        Processing time (w/o IO): 28.827736ms
        Processing time (w/o IO): 29.510347ms
        Processing time (w/o IO): 29.127541ms
        Processing time (w/o IO): 28.807536ms
        Processing time (w/o IO): 29.409245ms
        Processing time (w/o IO): 28.949638ms
        Processing time (w/o IO): 28.931938ms
        Processing time (w/o IO): 28.957138ms
        Processing time (w/o IO): 28.799236ms
        Processing time (w/o IO): 28.853737ms
        Processing time (w/o IO): 28.965739ms
        Processing time (w/o IO): 28.886338ms
        Processing time (w/o IO): 28.906238ms
          Time (mean ± σ):      69.1 ms ±   0.7 ms    [User: 62.2 ms, System: 13.2 ms]
          Range (min … max):    68.2 ms …  70.5 ms    10 runs
         
Go:

        Benchmark 1: ./related
        Processing time (w/o IO): 436.162085ms
        Processing time (w/o IO): 434.575667ms
        Processing time (w/o IO): 432.86844ms
          Time (mean ± σ):     578.4 ms ±   0.4 ms    [User: 575.8 ms, System: 31.7 ms]
          Range (min … max):   578.1 ms … 578.7 ms    2 runs
         
Go:

        Benchmark 1: ./related
        Processing time (w/o IO): 3.826478244s
        Processing time (w/o IO): 3.830682611s
        Processing time (w/o IO): 3.832434699s
          Time (mean ± σ):      4.324 s ±  0.019 s    [User: 4.352 s, System: 0.121 s]
          Range (min … max):    4.311 s …  4.338 s    2 runs

```

**Comment by DrBlury** [comment]
Okay I see the results but in theory it should be faster due to the more efficient mechanism for maintaining the top 5 list (less shifting of elements). And on my machine as well. Are you sure it's not just a VM spasm that is going evil against me?

**Comment by jinyus** [comment]
Are you using linux? It's very weird indeed. But the results of my machine matches the perf degradation on the VM. 

Did you run using docker? I can write a smaller docker file just for go if you don't want to build the big image.

**Comment by DrBlury** [comment]
I'm using MacOS (M1) here. I ran using `./run.sh go` so it's ARM.

**Comment by DrBlury** [comment]
You could add a minimal dockerfile for go that uses a base image which includes ARM in it's supported platforms. That would make it possible to test in docker.

**Comment by jinyus** [comment]
That seems to be the differentiating factor. Just substituting your new version without the other changes results in it being 5-6ms slower.

changing:

```go
minTags := isize(0) // Updated initialization

for j, count := range taggedPostCount {
	if count > minTags {
		// Find the position to insert
		pos := 4
		for pos >= 0 && top5[pos].SharedTags < count {
			pos--
		}
		pos++

		// Shift and insert
		if pos < 4 {
			copy(top5[pos+1:], top5[pos:4])
		}

		top5[pos] = PostWithSharedTags{Post: isize(j), SharedTags: count}
		minTags = top5[4].SharedTags
	}
}
```

to:

```go
for j, count := range taggedPostCount {
	if count > top5[0].SharedTags {
		top5[0] = PostWithSharedTags{Post: isize(j), SharedTags: count}
		for k := 0; k < topN-1; k++ {
			if top5[k].SharedTags > top5[k+1].SharedTags {
				top5[k], top5[k+1] = top5[k+1], top5[k]
			}
		}
	}
}
```

jinyus:
They are equal on linux arm. The new version is ~1ms faster:

old:

```
Benchmark 1: ./related
Processing time (w/o IO): 38.729717ms
Processing time (w/o IO): 38.882679ms
Processing time (w/o IO): 38.532477ms
Processing time (w/o IO): 38.519756ms
Processing time (w/o IO): 39.1598ms
Processing time (w/o IO): 38.998638ms
Processing time (w/o IO): 38.982198ms
Processing time (w/o IO): 39.485761ms
Processing time (w/o IO): 38.896917ms
Processing time (w/o IO): 39.178919ms
Processing time (w/o IO): 38.907718ms
Processing time (w/o IO): 38.608556ms
Processing time (w/o IO): 38.779797ms

```

new:

```
Benchmark 1: ./related
Processing time (w/o IO): 38.006301ms
Processing time (w/o IO): 37.994581ms
Processing time (w/o IO): 37.846461ms
Processing time (w/o IO): 37.72482ms
Processing time (w/o IO): 37.835541ms
Processing time (w/o IO): 38.165302ms
Processing time (w/o IO): 38.382104ms
Processing time (w/o IO): 37.987501ms
Processing time (w/o IO): 38.336144ms
Processing time (w/o IO): 38.221782ms
Processing time (w/o IO): 37.938342ms
Processing time (w/o IO): 37.949861ms
Processing time (w/o IO): 37.693219ms
```

DrBlury:
Wow that's interesting. I wonder why arm is faster here. What exactly is different in the generated machine code?

jinyus:
Link to the [compiler output](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DBUpJfWQE8Ayo3QBhVLQCuLBiADMpOwBk8DTAA5VwAjTGIQACZI0gAHVAVCawZHFzdPOISkgT8A4JYwiOizTAsrASECJmICVNd3L3NMS2TK6oJcoNDwqJiFKpq69MaBjv8ugp7ogEozVGdiZHYOWKZkAGsmYEwAahYmfwBSDQBBY5OCAE9Y3eUEggB1QgQhBGrMdAAVLYUd/uJnJYdocAOwAIXOxwAnHd%2Bjt4Qidv4CJCNFDXu8vj8kYJISCACKok5UZwMZB7A4MCDTYHg1FQhiuX4gQ4efHAgCsEI5%2BORoLBXFIOxiOy8O2kOw5QoAbEKQUKABxCqFCrgaUGE070oixDk7FlszlgjmHHmwx7PDHED7fYAKfkaonQ/jEHbaIVoUkEfWs9nEQzbHaMli/B1atHQvBUHaewTAjzYHY6k1c9U8gB0Vpt2LDZwj%2BahydNYLT7N9O3NTwILze1qxdv55pAbqFWfrCmbsZRBPp0OdJB26x9hvVHjBg/j9klAFouKzx%2BtDpEIcvaRDw1C%2BwWo0nULri4uM23baGE7v91zF8u50fa9m7WvewWt4W9ymwYf8UKi5elwLTWWho/h%2Bf43l%2B57vp%2BT6bs%2BPYbi%2BjrwRGcF5puKGIScHCzLQnAcrw7gcFopCoJwADiqB/PMiy7EuHg8KQBCaFhszrL06aSNKHIghyHiSJEHFcTxUo4Rwkj4UxxGcLwHYaAxTGzHAsBIGgLCxHQ4TkJQKlqfQETVCw0qSAA%2BsAtBWLEpBYAAbngSwAGp4JgADuADyNwEfRNBmeEHYQCEEkhP41SXJw9GBcwxCXC5ITaM0jHcLwKlsIILkMLQIWEbwWAhM4wD2GItAdgllmYPsRjiJlll4NaLRWZgRVEZgqjNM4BDLPRyKlBJtB4CE/qRY4WASQQxB4CwoVYXwBh2g5zluYwE0yIIIhiOwUhLfIShqBJuiCgYRggCY%2Bi9R2kCzHu5QMEVvCoHVxCjVgp3UiUZTJLYDAOE49R6L44z5IUejxIkl1DO4gpA9kDCdP9PSCk0LQVKMoN6PDl1tDU0PdBEcNI196Q4%2B0mOTNjswKFRSwSNhuHiZVJEcDsJwAEoALKGTswDkhA5k0hAuCEAOtFcNMvDxVo0wsSAkiSOmkQaGiXAePxsvy14IliaQBFEXT0kgLJouTYpiAoKgqnqWQFAQNpZucB4CrSiZZl4LEWWYDZ9mOa57mLV5bXEL5/mVeFwWLUHkXRbFliLUljAEKl6USdluX5bQhWLVgZWmcsRH4DVVh1Q1vBNS1bWLZ1IlET1fXBYNWci6N40JbMVDTQos2ewtxX8MtojiOtXebSo6iVboMT7cYpiV0952xJd13EXdD31fApOlHFb0QHYyOCr9eRY4DWQg3jYOZMDyREwDcOrwjDDo7UR8o1faOjOfsNmLjaTH/0hN/XvQtzAsFM/4iTwhrCSdMdiqDtuzTm3Mdi83wEQV0gthZyUyuLUgrEPAeHTFg3BeD8H6E4OrTWN0pJmF1qgsWVMOCRBplrMh%2Bt0F3USDYSQQA%3D%3D%3D)

PR #212, opened 2023-10-12
Update app.d

Imperatorn:
I guess the output filename should be _con?

### DIFF
--- a/d_con/source/app.d
+++ b/d_con/source/app.d
@@ -98,5 +98,5 @@ void main()
 	sw.stop();
 
 	writeln("Processing time (w/o IO): ", sw.peek.total!"usecs" * 1.0 / 1000, "ms");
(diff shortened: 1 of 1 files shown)


cyrusmsk:
I also would like to make small changes into this file, before merging.

Imperatorn:
> I also would like to make small changes into this file, before merging.

Sorry, I also forgot to check the output file, did a separate PR. Sorry for the mess

cyrusmsk:
> I also would like to make small changes into this file, before merging.

ok. I've prepared new PR with small adjustments.

cyrusmsk:
Can you remove "result" array from the code - it is doing nothing

Imperatorn:
> Can you remove "result" array from the code - it is doing nothing

Looks like I missed a lot of things today. Will fix

PR #213, opened 2023-10-12
Update F# to dotnet 8, add fsharp AOT version

delneg:
https://github.com/jinyus/related_post_gen/pull/198
After C#'s upgrade to dotnet 8 & AOT, it's only fair to do the same for F#

Also, I've added ServerGarbageCollection to match C# version exactly

### DIFF
--- a/fsharp/Program.fs
+++ b/fsharp/Program.fs
@@ -15,7 +15,6 @@ type Post =
       title: string
       tags: string[] }
    
-[<Struct>]
 type RelatedPosts =
     { _id: string
       tags: string[]
--- a/fsharp/Program.fs
+++ b/fsharp/Program.fs
@@ -2,8 +2,7 @@
 open System.IO
 open FSharp.NativeInterop
 open System.Collections.Generic
-open System.Text.Json
-
+open FSharp.Json
 #nowarn "9"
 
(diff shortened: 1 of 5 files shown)


delneg:
I've tested now, and looks like AOT is broken for F#.
Trying to fix it, but JSON serialization is still broken

delneg:
Yeah, looks like I've fixed it.
For me it got faster from 32ms to 28-29ms locally

delneg:
Ok, so let me explain the changes made:
1) Had to add .rd.xml file which instructs ILC to not link out the necessary types during AOT compilation
2) Switched JSON ser/de from System.Text.Json (which is not AOT compatible due to sourcegen) for FSharp.Json
3) Update to dotnet 8
4) Enable server GC like in C#
5) add AOT mode to run.sh


I've removed [<Struct>] for RelatedPosts to not impact any performance

jinyus:
What do you use to install dotnet, only v7 is in the arch repo. I need to update the dockerfile

delneg:
You can find installer & binaries here
https://dotnet.microsoft.com/en-us/download/dotnet/8.0

delneg:
Worked for me:
```
wget https://dot.net/v1/dotnet-install.sh -O dotnet-install.sh
chmod +x ./dotnet-install.sh
./dotnet-install.sh --channel 8.0
```

PR #215, opened 2023-10-12
Cleanup

Imperatorn:


Imperatorn:
Is it ok to reuse related_posts_d.json as output?

jinyus:
no, use a separate file. I will update it

PR #221, opened 2023-10-13
[CSharp] Remove stackalloc

MichalStrehovsky:
This is a pretty bad deoptimization. Seeing 68 ms on main vs 38 with this PR.


jinyus:
Seeing a smalller difference, about ~1ms on my machine. `34ms -> 33ms`

PR #223, opened 2023-10-13
[Julia] Use an ad hoc package: SuperDataStructures.jl.

LilithHafner:
You should probably double-check that this is within the rules. I made a custom package just for this benchmark. It's https://github.com/LilithHafner/Jokes/tree/main/SuperDataStructures.jl. This PR uses the data structures defined there to get significant speedups.

Julia is a cool language in that it allows unbounded specialization and speculative optimization within any given API, which allows an upstream package (e.g. SuperDataSructures) to make types that are quite efficient in a given downstream usecase (e.g. this)

I want to be totally clear that while SuperDataStructures.jl does provide SuperVector and SuperDict which are fully functional types that fulfill the AbstractVector and (write only) AbstractDict APIs, it also contains optimizations tailored very specifically to the code in this PR. It would be accurate to say that those optimizations are "snuck in". They make this code run way faster than it has any reasonable right to run and the compiled code will look nothing at all like the surface syntax, but the optimizations do not reduce correctness at all (except for bugs—it's typically safe to assume there are a few bugs).

The upstream package does not use any inbounds, ffi, unsafe, etc.


jinyus:
Nice hack but I won't allow it.

LilithHafner:
What rule(s) does it break? / is there any particular reason you won't allow it?

jinyus:
Contributions to the competition are based on good faith, so if you decide that caching the results during the warmup run falls within the bounds of the current rules. You should know that this goes against the spirit of the competition. A simple change to the size of the array used in the warmup run and your hack falls apart. It also assumes that the array contains a least 5 items which is also against the rules.

```julia
function relatedIO()
    json_string = read("../posts.json", String)

    related(posts[1:10])

    json_string = read("../posts.json", String)
    posts2 = JSON3.read(json_string, Vector{PostData})

    start = now()
    all_related_posts = related(posts2)
    println("Processing time (w/o IO): $(now() - start)")


    open("../related_posts_julia.json", "w") do f
        JSON3.write(f, all_related_posts)
    end
end
```

LilithHafner:
> caching the results during the warmup run

Yikes, that would be super duper icky. I don't do that. I'll remove the warmup to make it clear that's not the case.

> A simple change to the size of the array used in the warmup run and your hack falls apart

That is bad; unfortunately, I'm not able to reproduce that locally. When I run the code you posted it reports a fast runtime. Would you give me more specific instructions for how to reproduce this problem?

> It also assumes that the array contains a least 5 items which is also against the rules.

What should I return as the 5 most closely related posts if there are fewer than 5 posts total? I don't see how to avoid this assumption.

> ps: the rules were never static to begin with. I have to adjust them from time to time

I support this.

LilithHafner:
I deleted the warmup in the [source branch of this PR](https://github.com/LilithHafner/related_post_gen/tree/lh/julia-super-opt), the change is not propagating to here because the PR is closed. The diff is just

```diff
diff --git a/julia/related.jl b/julia/related.jl
index d3d840e..366dfad 100644
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -10,7 +10,6 @@ function relatedIO()
     json_string = read("../posts.json", String)
     posts = JSON3.read(json_string, Vector{PostData})
 
-    related(posts)
 
     start = now()
     all_related_posts = related(posts)
```

jinyus:
Majority of the work is being done outside the timer:

new: 112ms 
old: 91ms

```julia
function relatedIO()
    json_string = read("../posts.json", String)


    all_related_posts = related(posts)
    start = now()
    
    open("../related_posts_julia.json", "w") do f
        JSON3.write(f, all_related_posts)
    end
    println("Processing time (w/o IO): $(now() - start)")
end
```

LilithHafner:
That work is almost all compilation time. The fact that it increased is strange, as the value and type of `all_related_posts` is unchanged by this PR. I've added a warmup for writing to disk which cuts it way down (pushed to source branch). The warmup runs before the program has access to any of the input data. If you're still not convinced that all the non-IO work happens in the `related` function, no worries, just time it with IO included. JSON3.jl is fast.

<details>
<summary>diff</summary>

```diff
diff --git a/julia/related.jl b/julia/related.jl
index 366dfad..da34313 100644
--- a/julia/related.jl
+++ b/julia/related.jl
@@ -4,21 +4,21 @@ using Dates
 using StaticArrays
 using SuperDataStructures
 
-# warmup is done by hyperfine
+function write(filename, data)
+    open(filename, "w") do f
+        JSON3.write(f, data)
+    end
+end
 
 function relatedIO()
     json_string = read("../posts.json", String)
     posts = JSON3.read(json_string, Vector{PostData})
 
-
     start = now()
     all_related_posts = related(posts)
     println("Processing time (w/o IO): $(now() - start)")
 
-
-    open("../related_posts_julia.json", "w") do f
-        JSON3.write(f, all_related_posts)
-    end
+    write("../related_posts_julia.json", all_related_posts)
 end
 
 struct PostData
@@ -67,4 +67,7 @@ function related(posts)
     relatedposts
 end
 
+# warmup for writing JSON to disk
+write(joinpath(tempdir(), "warmup_related_posts.json"), [RelatedPost("a", ["b"], fill(PostData("c", "d", ["e"]), 5))])
+
 const res = relatedIO()
```

</details>

jinyus:
Not seeing much difference now:

```
Benchmark 1: julia related.jl
Processing time (w/o IO): 21 milliseconds
Processing time (w/o IO): 26 milliseconds
Processing time (w/o IO): 30 milliseconds
Processing time (w/o IO): 21 milliseconds
Processing time (w/o IO): 29 milliseconds
Processing time (w/o IO): 27 milliseconds
Processing time (w/o IO): 21 milliseconds
  Time (mean ± σ):      3.078 s ±  0.065 s    [User: 2.919 s, System: 0.159 s]
  Range (min … max):    2.986 s …  3.198 s    10 runs
```

LilithHafner:
Lovely, I'm assuming those timings include (IO). If this is within the rules, even if you don't think it's much faster, I'd love to see it merge. We can work on tuning the performance further in later PRs. I've adjusted the benchmarking to include IO times in the source branch.

The part of this that is questionable in my opinion is the use of speculative optimization, lazy evaluations, and broadcast fusion to achieve macro-optimizations that are not possible without such tricks. None of those things extend above the level of the `related()` call, though.

LilithHafner:
Friendly bump, is there any reason remaining why you are holding back on merging?

According to your benchmarks, it's not a big speedup, but it is still an improvement.

jinyus:
rename it to julia_super. I don't want to replace the current solution with this.

jinyus:
I''m not sure if you're ready to merge, so I'll put this in drafts until you're ready

julia_highly_optimized/related.jl
```diff
@@ -0,0 +1,73 @@
...
+end
+
+function relatedIO()
+    start = now()
+    json_string = read("../posts.json", String)
+    posts = JSON3.read(json_string, Vector{PostData})
+
+    all_related_posts = related(posts)
+
+    write("../related_posts_julia_highly_optimized.json", all_related_posts)
```
jinyus, line 20: I/O shouldn't be measured. I was just suspicious.

LilithHafner:
I'm ready.

LilithHafner:
I like what you've done putting Julia HO at the top with a disclaimer but still highlighting in green the fastest "sensible" runtime. It straddles a nice compromise between recognizing that this does the work and follows the rules and also recognizing that it isn't exactly a "normal" approach.

PR #224, opened 2023-10-13
speed up lookups using MVector

arikheinss:
Hello,

I replaced two local Variables in the standard julia implementation which were Vectors with MVectors, which pushed the performance on my machine from ~24ms to ~19ms.

Excuse the needless merge/unmerge commits, I am somewhat inexperienced with Github and have not figured out the workflow perfectly.

### DIFF
--- a/crystal/src/crystal.cr
+++ b/crystal/src/crystal.cr
@@ -38,12 +38,11 @@ posts.each_with_index do |post, i|
 end
 
 allRelatedPosts = Array(RelatedPost?).new(posts.size, nil)
-# using Pointer buffer can greatly improve performance but this is unsafe, thus against the benchmark rules
-# https://crystal-lang.org/reference/latest/syntax_and_semantics/unsafe.html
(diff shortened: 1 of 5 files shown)


jinyus:
> I replaced two local Variables in the standard julia implementation which were Vectors with MVectors

This has been discussed #194 and MVectors are banned as it uses unsafe code for direct memory access, which is against the [rules](https://github.com/jinyus/related_post_gen#run-benchmark).


> Excuse the needless merge/unmerge commits, I am somewhat inexperienced with Github and have not figured out the workflow perfectly.

That's fine, I will just cherry pick the commits that I want to merge. Just make smaller commits to make my job easier.

I see you made some changes to the crystal code? Did you want to include those in the PR?

arikheinss:
Nope, I just modified the julia code and only in 3 lines (the two variables, changing a hardcoded variable to a global constant, and one extra line where my LSP formatter removed some trailing whitespace). All this is in the first commit, that mentions the MVectors.

All the other stuff that changed comes because I did an accidental pull before the push and do not know how to get this out of my history now. If MVectors are banned then I have nothing to contribute.

PR #230, opened 2023-10-13
Update app.d

Imperatorn:
Some formatting and updated algo

### DIFF
--- a/d/source/app.d
+++ b/d/source/app.d
@@ -60,9 +60,7 @@ void main()
 			{
 				int upperBound = TopN - 2;
 
-				ubyte sharedTagsCount = top5[upperBound].sharedTags;
-
-				while (upperBound >= 0 && count > sharedTagsCount)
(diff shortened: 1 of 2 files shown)


Imperatorn:
On my local machine with latest Go and D, D performs slightly better than Go.

Will be interesting to see if you get the same results.

jinyus:
Faster but not correct:

```
Checking output
Error: Post jf0da9ku is invalid!
expected: 15 shared tag count
actual: 11
```

This means it's not selecting the correct top5

Imperatorn:
Oh, I messed up. Will fix, thanks

Imperatorn:
> Faster but not correct:
> 
> ```
> Checking output
> Error: Post jf0da9ku is invalid!
> expected: 15 shared tag count
> actual: 11
> ```
> 
> This means it's not selecting the correct top5

Fixed the error

jinyus:
~1ms faster

PR #234, opened 2023-10-13
Lookup table

Imperatorn:
Version with lookuptable and newaa


Imperatorn:
Wait a bit with merging this. I'm running some tests locally. It should be ok, but I just want to be sure.

Imperatorn:
Removed the .ptr as per suggestion it might be against the rules to use it.

PR #239, opened 2023-10-14
[C#] Use vectors to find top posts

MichalStrehovsky:
These are lowered to whatever is the platform-specific vector (and whatever is the platform-specific vector size).

To be 100% clear, the code in the benchmark is not hardware specific: the vector sizes are determined at compile- (AOT) or run-time (JIT) based on CPU capabilities and not hardcoded or special cased in the source code. .NET has a crossplat vector abstraction.

### DIFF
--- a/csharp/Program.cs
+++ b/csharp/Program.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using System.Numerics;
 using System.Runtime.InteropServices;
 using System.Text.Json;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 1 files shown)


delneg:
Basically, Vectors are SIMD-enabled
I doubt that Vectors are "general purpose data structures"
Sure, they're included in stdlib, but I doubt someone would use them for "normal" usecase
Also, IMO it lies too close to "hardware-specific" boundary @jinyus

https://learn.microsoft.com/en-us/dotnet/standard/simd

MichalStrehovsky:
Like the linked docs say, processing data in vectors is common in mathematical apps and this is a compute benchmark.

Many compilers will parallelize as part of optimizations. For example, VC++ will do this. It's not magic, one just needs to [inspect logs](https://learn.microsoft.com/en-us/cpp/build/reference/qpar-report-auto-parallelizer-reporting-level?view=msvc-170) and put the code into exact shape that can be parallelized (it tends to be very fragile/finnicky, but it works once you get the shape right). In .NET it's more explicit. Unless we forbid people to use logs to ensure their code gets autovectorized (if the language itself doesn't provide primitives for guaranteed parallelization like .NET), using this just gives C# an even ground. Removing will give autovectorizers an advantage - but like I said - they're not magic, one needs to often consult the logs to ensure it kicks in, that work is just not visible.

cyrusmsk:
I agree that clarifications if SIMD-backed vectors are allowed are needed.
Because others could use them in this case as well:
* https://ziglang.org/documentation/master/#Vectors
* https://dlang.org/spec/simd.html
* Rust has a couple of libraries (for example this one https://docs.rs/numeric-array/latest/numeric_array/)

jinyus:
I going to modify the rule to only prevent SIMD in single threaded solutions.

zigzag312:
I'm leaning towards no SIMD-backed vectors. Flow of C# code with vectors is harder to follow, and it's not something I would do for normal data processing as it costs more developer time and is harder to read. I would usually do something like that only when I can afford to spend time on max optimization. 

Autovectorizers can sometimes work even when there is no extra effort put in from the developers side. But I agree it creates a bit of unfair situation, if someone will devote extra time to put the code into exact shape that can be autovectorized. 

Rule to allow SIMD only in multi-threaded solutions seems good to me. Maybe also add rule to prevent very non-idiomatic code for single threaded solutions, to prevent very unusual code shaping just to get it autovectorized.

PR #244, opened 2023-10-14
[rust] Some minor optimizations

cloneable:
* Switch tag indices to u8 like D has.
* Make Post more cache line friendly.
* Use SmallVec to store str references inline.
* rust_con: Use local ThreadPool and initialize post-count TLS.
* Some cleanups.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -113,30 +113,19 @@ dependencies = [
 
 [[package]]
 name = "errno"
-version = "0.3.3"
+version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "136526188508e25c6fef639d7927dfb3e0e3084488bf202267829cf7fc23dbdd"
+checksum = "ac3e13f66a2f95e32a39eaa81f6b95d42878ca0e1db0c7543723dfe12557e860"
 dependencies = [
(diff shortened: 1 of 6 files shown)


cyrusmsk:
Btw can we remove this "mimalloc" thing from Rust code?
I suppose almost every compiled language could make some changes in default allocators settings. But except Rust nobody is doing it.
Of course mimalloc is a great thing, but I don't think it should be applied only for Rust
@jinyus

cyrusmsk:
> Move tag lookup table creation before counter like other languages.

This was also fixed in D. Now lookup table created after the counter.

cyrusmsk:
Thanks!

jinyus:
It seems smallvec invokes unsafe code: https://docs.rs/smallvec/latest/src/smallvec/lib.rs.html#7-2463

cloneable:
Lots of things contain unsafe code. There's nothing wrong with unsafe code, provided UB soundness is checked. smallvec is a very popular crate. But I can understand why the benchmark code itself should not resort to unsafe code.

jinyus:
That's a slippery slope. It's better to ban unsafe code outright.

jinyus:
wish you have made smaller commits so I can cherry pick them. I will try to merge manually.

PR #247, opened 2023-10-14
Initialization after timer

cyrusmsk:
Checked zig, go and Julia.
Move data initialization after timer started.


Imperatorn:
Good catch

PR #249, opened 2023-10-14
Revert "[Nim] compile with clang"

jinyus:
Reverts jinyus/related_post_gen#237

Performance regression: #248


michaelsbradleyjr:
Something is a bit strange/interesting, then, because on my laptop (2019 macbook pro x86_64) the times went very consistently  from roughly 40 ms to 20 ms, i.e. big performance increase.

michaelsbradleyjr:
I don't see a regression in #248, instead the Nim results seem to be missing:

https://github.com/jinyus/related_post_gen/commit/d036bba361cf21dda68395126498a5700b4541ce

https://github.com/jinyus/related_post_gen/blob/d036bba361cf21dda68395126498a5700b4541ce/raw_results.md

PR #250, opened 2023-10-14
[Nim] clang again

inv2004:
I found that prev PR about Nim changed full structure of the nimble project. I applied some changes (which did not improve time), but to make it a bit more maintable

cc @michaelsbradleyjr

### DIFF
--- a/nim/related.nimble
+++ b/nim/related.nimble
@@ -13,3 +13,11 @@ bin           = @["related"]
 requires "nim >= 2.0.0"
 requires "jsony"
 requires "xxhash"
+
+task buildopt, "build optimized":
+  exec "nimble -d:profileGen -d:release build"
+  echo "Gen stat"
+  exec "./related > /dev/null"
+  exec "llvm-profdata-12 merge default.profraw --output default.profdata"
+  exec "nimble -d:profileUse -d:release build"
+
--- a/nim/related.nimble
+++ b/nim/related.nimble
@@ -18,6 +18,6 @@ task buildopt, "build optimized":
   exec "nimble -d:profileGen -d:release build"
   echo "Gen stat"
   exec "./related > /dev/null"
-  exec "llvm-profdata-12 merge default.profraw --output default.profdata"
+  exec "llvm-profdata merge default.profraw --output default.profdata"
   exec "nimble -d:profileUse -d:release build"
 
--- a/nim/config.nims
+++ b/nim/config.nims
@@ -1,4 +1,5 @@
 switch("opt", "speed")
-switch("passC", "-flto  -Ofast")
-switch("passL", "-flto  -Ofast")
+switch("passC", "-flto -Ofast")
+switch("passL", "-flto -Ofast")
+switch("cc", "clang")
 switch("panics", "off")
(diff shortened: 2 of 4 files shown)


nim/related.nimble
```diff
@@ -13,3 +13,11 @@ bin           = @["related"]
 requires "nim >= 2.0.0"
 requires "jsony"
 requires "xxhash"
+
+task buildopt, "build optimized":
+  exec "nimble -d:profileGen -d:release build"
+  echo "Gen stat"
+  exec "./related > /dev/null"
+  exec "llvm-profdata-12 merge default.profraw --output default.profdata"
```
jinyus, line 21: is llvm12 needed

inv2004: Sorry, it is just my old ubuntu, will fix in a minute

michaelsbradleyjr:
@inv2004 that's fine. Fwiw, my changes weren't made on a whim. When working on this, I'm mainly in the shell on macOS, test in Docker in an Ubuntu container, and try things out in a GHA run on my fork. So some of the changes were to make it simpler to work across platforms without doing anything particularly weird.

I've lost any love I had for nimble and generally avoid nimble tasks, instead doing thing in terms of `nim c`, `nim.cfg` and/or `config.nims` (plug `-d:` opts and env vars). That seems to be the way things are going with [`atlas`](https://github.com/nim-lang/atlas) on the horizon, but it's a stylistic preference at this point, so whatever works I guess.

In `config.nims` I had setup `nim/nimcache` relative to the root of this project because across 5k ... 60k runs llvm's profiler was exiting with error if nimcache wasn't cleaned out. In those kinds of situations, I find it's nicer to have a project-specific nimcache than munging around in `${HOME}/.cache/nim/`. I'll check in a GHA run on my fork if the changes you introduced in this PR run into the same problem and will submit a one-or-two-liner fix if that's the case.

Finally, I'm a bit confused about the revert of my #237. It was to address a perf regression in #248, but if you look at the commit/results for #248 (which was reverted along with #237 in #249), the Nim results are completely missing. See https://github.com/jinyus/related_post_gen/commit/d036bba361cf21dda68395126498a5700b4541ce and [d036bba/raw_results.md](https://github.com/jinyus/related_post_gen/blob/d036bba361cf21dda68395126498a5700b4541ce/raw_results.md). So was there a regression, or something just got screwed up in the run, or...?

michaelsbradleyjr:
I figured out a couple of subtle things with 237. I'm making these notes here in case someone might find them useful, but mostly just because I'm happy I solved the problem.

* I was using an Ubuntu container instead of Arch 🤦 . I should have been using the same container setup per the Dockerfile in this project, but I hadn't looked at it and had in mind the GitHub Actions environment. That led to a mismatch in the name of the llvm-profdata executable, which causes a command to fail, and that's probably why the Nim data was missing in 248.
* clang 15/earlier's LLVM gold plugin seems tolerant of subtle changes between the profiled/optimized objects that can cause clang 16 to fail across repeat runs. Also in clang15/earlier there's a noticeable perf loss on Linux *but not* on macOS.

While I don't intend them for a PR, the simple fixes can be seen here: [fixes](https://github.com/michaelsbradleyjr/related_post_gen/compare/nim_rev_3...michaelsbradleyjr:related_post_gen:nim_rev_3_fixes).

With those changes, the performance gap between this PR and 237 is closed.

cyrusmsk:
Btw are you on x86 Mac?

michaelsbradleyjr:
> Btw are you on x86 Mac?

Yes.

PR #251, opened 2023-10-14
[cpp con] add cpp concurrent

kaiwu:
first cpp concurrent version


cyrusmsk:
this looks like a "specific adjustment" :)
`char padding[2 * CACHE_LINE_SIZE - 88];`

kaiwu:
it's not, cpp needs to manually adjust struct size to be multiple of cache line size for concurrent processing, it is to avoid false sharing

btw please use clang++ for concurrent

cyrusmsk:
C++ code currently is not assigning top5 posts into resulted array (relatedPosts). Please check @jinyus

jinyus:
`relatedPost.related[upperBound + 1] = &posts[j];`

seems like an optimization.

@cyrusmsk

cyrusmsk:
Oh it is using references. I haven’t noticed that. Thnx

cloneable:
fyi, this is unsafe: `posts[j]`

PR #255, opened 2023-10-14
Update app.d

Imperatorn:
Disable GC. Is this cheating? 

We do that quite often in D if we have some code that is short lived, typical example is actually something like your example.


jinyus:
Let's keep it on since it doesn't make a difference and would be unfair to gc languages.

PR #259, opened 2023-10-14
Update app.d

Imperatorn:
Don't do unnecessary lookup


jinyus:
looks ugly but saves 1ms

PR #261, opened 2023-10-15
[Nim] PGO can generates bad stats. Workaround to reduce it

inv2004:
Looks like PGO can hit or miss collect "good" stats. 50%/50% for me.

I do not know why it is possible, but generating more stats helps (not 100%)

Here is "good" result:
image

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -35,7 +35,11 @@ if command -v hyperfine &>/dev/null; then
     HYPER=1
 fi
 
-time="$(which gtime 2>/dev/null || which time)"
+time="$(which gtime 2>/dev/null || which time 2>/dev/null)"
+if [[ -z "${time}" && -z "$(which hyperfine 2>/dev/null)" ]]; then
+  echo 'time, gtime, or hyperfine must be available in $PATH'
+  exit 1
+fi
 
 # capture the output of a command and write it to stout or to a file
 capture() {
--- a/run.sh
+++ b/run.sh
@@ -35,7 +35,7 @@ if command -v hyperfine &>/dev/null; then
     HYPER=1
 fi
 
-time="$(which gtime || which time)"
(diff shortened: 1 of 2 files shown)


inv2004:
You can find the problem on the picture:

laptop is with external power, no even browser running

image

run.sh
```diff
@@ -35,7 +35,7 @@ if command -v hyperfine &>/dev/null; then
     HYPER=1
 fi
 
-time="$(which gtime || which time)"
+time="$(which gtime 2>/dev/null || which time)"
```
michaelsbradleyjr, line 38:
```suggestion
time="$(which gtime 2>/dev/null || which time 2>/dev/null)"
if [[ -z "${time}" && -z "$(which hyperfing 2>/dev/null)" ]]; then
  echo 'time, gtime, or hyperfine must be available in $PATH'
  exit 1
fi
```

Might as well change it for both and exit with error if both are missing and `hyperfine` is missing.

GNU `time` isn't installed by default in e.g. `archlinux:base`. The shell built-in may be available, but that's not the same thing and isn't equivalent. I noticed that today when I started using an Arch container instead of Ubuntu.

michaelsbradleyjr:
Even with `STEP_COUNT = 100` or `STEP_COUNT = 1000` and several runs with 10/100/100, I didn't see any performance gains on my mac laptop, whether building in macOS (clang 15) or in an Arch container (clang 16).

So I think it's a good find by @inv2004, but if it's really that hit-and-miss and may vary by machine, I wonder if it's better to just leave out the `STEP_COUNT` changes. A comment could be added to `related.nimble` explaining that the PGO doesn't give deterministic results.

inv2004:
@michaelsbradleyjr try to switch to gcc + just build -d:release. I found that it generates +1ms from "good" PGO, but with stable result.

PR #266, opened 2023-10-15
[Crystal] Improve time measurement

ilevd:
More correct time measurement with small changes in hashmap init capacity.


jinyus:
the old tagmap init is faster so I kept it.

PR #268, opened 2023-10-15
[cpp] improve on large dataset

kaiwu:


cyrusmsk:
> *(taggedPostCount.data() + otherPostIdx) += 1;
is it sage doing like that instead of 
> taggedPostCount[otherPostIdx] += 1;

cyrusmsk:
C++ vector probably could use "at" instead:
https://cplusplus.com/reference/vector/vector/at/

>The function automatically checks whether n is within the bounds of valid elements in the vector exception if it is not (i.e., if n >is greater than, or equal to, its size. This is in contrast with member operator[], that does not check against bounds.

kaiwu:
with enough optimization level `[ ]` operator is no different as pointer de-reference, the major change of the PR is that for large data set `int` is a lot faster than `size_t`

kaiwu:
the Zig implementation suffers the same issue, which will be dealt with PR separately

cyrusmsk:
Probably [] and pointer de-reference are the same :) But you need to use "at" to be compliant with the Rules

kaiwu:
by all means rules are rules, but performant code avoid every unnecessary check since it's made valid in previous setups.

cyrusmsk:
The rules are stating clearly that "Disabling runtime checks (bounds etc)" is not allowed.
C++ vector is able to use bounds-checking and production code should always do that.

PR #270, opened 2023-10-15
add Erlang bench

NoobsEnslaver:
Now it's not so fast, but still.
If need add something else - just let me know


jinyus:
Looks good.

PR #276, opened 2023-10-15
[Clojure] Add implementation

ilevd:
First implementation.


clojure/src/related/core.clj
```diff
@@ -0,0 +1,74 @@
...
+                                                                         (recur (- upper-bound 2)))))]
+                                                            (aset top5 (+ up 2) cnt)
+                                                            (aset top5 (+ up 3) i)
+                                                            (recur (inc i) (aget top5 8)))
+                                                          (recur (inc i) min-tags)))))
+
+                                                  {:_id     (:_id post)
+                                                   :tags    (:tags post)
+                                                   :related [(nth posts (aget top5 1))
+                                                             (nth posts (aget top5 3))
```
jinyus, line 63: Manually indexing isn't allowed. Use integer division

```js
for (let i = 1; i < 10; i += 2) {
      topPosts[Math.floor(i / 2)] = posts[top5[i]];
}
```

or skip and append:

```crystal
(1...10).step(2) do |i|
    top_posts << posts[top5[i]]
end
```

ilevd: Why not? It's just a little optimization for not to create `range` or loop with its vars.

jinyus:
Very slow for me so I'll exclude from the charts for now, as it'd be misleading.

```
Processing time (w/o IO): 162077
```

btw a unit (s/seconds/ms/milliseconds) should there.

clojure/src/related/core.clj
ilevd: So what exact rule this technique against? :)

jinyus: I didn't think of making a rule for this as it's a bad practice in general. I should be able to change 1 constant and get the topN without looking at the code. I know some of the solutions doesn't adhere to this but that could be done with a slight medication. 

ie: changing the 10 above to 2*topN

ilevd: I think it depends. It may be considering as an array initialization like e.g in Zig:

```
var top_5 = [5]u8{ 0, 0, 0, 0, 0 };
```

Anyway, it would be better to make some note in the rules, because making some optimizations and getting that it isn't allowed because it's a bad practice in general in the benchmark repo a little bit strange.

ilevd: I think it depends. It may be considering as an array initialization like e.g in Zig:
```
var top_5 = [5]u8{ 0, 0, 0, 0, 0 };
```
Anyway, it would be better to make some note in the rules, because making some optimizations and getting that it isn't allowed because it's a bad practice in general in the benchmark repo is a little bit strange.

PR #280, opened 2023-10-15
[Clojure] Improvements

ilevd:
Faster array getters/setters.


jinyus:
Much faster:

`Processing time (w/o IO): 3183ms`

PR #288, opened 2023-10-16
Speedup erlang solution

NoobsEnslaver:
was 9 sec on my pc, now 3 sec.
What erlang version do you use? Use 25+ plz.


jinyus:
rebar 3.22.0 on Erlang/OTP 26 Erts 14.1

PR #291, opened 2023-10-16
[Nim] concurrent

michaelsbradleyjr:
It works just fine, but performance relative to the single-threaded implementation seems to improve only with 8 or 16 cores/threads. I'll explain below, but for that reason I've made it a draft PR. If you're happy with it, feel free to take it out of draft and merge as is, and in any case I'll be working on improving the performance for 4 cores.

On my laptop with 16 cores and with 16/8 threads, I'm seeing processing times in the low teens. The numbers bounce around a bit across builds and runs, up into the 20-30 ms range, maybe owing to the inconsistent PGO discussed in #261 or maybe for some other reason/s. 

When I limit it to 4 threads, the performance is worse than the single-threaded Nim implementation. That's disappointing, but I'm optimistic there are ways to improve it.

---

With `-d:danger` builds and 16 threads I'm seeing processing times in the single digits: very nice to see! but not admissible per the rules, I know.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -491,10 +491,11 @@ run_nim() {
 run_nim_con() {
     echo "Running Nim Concurrent" &&
         cd ./nim_con &&
+        echo "using ${nproc} threads" &&
         nimble -y install -d &&
         ./buildopt.sh ${nproc} &&
         if [ $HYPER == 1 ]; then
-            capture "Nim Concurrent (nproc=${nproc})" hyperfine -r $runs -w $warmup --show-output "./build/related_con"
+            capture "Nim Concurrent" hyperfine -r $runs -w $warmup --show-output "./build/related_con"
         else
             command ${time} -f '%es %Mk' ./build/related_con
         fi
@@ -916,7 +917,7 @@ elif [ "$first_arg" = "all" ]; then
         run_java_graal || echo -e "\n" &&
         run_java_graal_con || echo -e "\n" &&
         run_nim || echo -e "\n" &&
-        run_nim_con || echo -e "\n" &&
+        #run_nim_con || echo -e "\n" && #too slow, excluded for now
         run_fsharp || echo -e "\n" &&
         run_fsharp_con || echo -e "\n" &&
         run_fsharp_con_aot || echo -e "\n" &&
--- a/run.sh
+++ b/run.sh
@@ -41,6 +41,12 @@ if [[ -z "${time}" && -z "$(which hyperfine 2>/dev/null)" ]]; then
     exit 1
 fi
 
+if [[ -n "$(which nproc 2>/dev/null)" ]]; then
+    nproc=$(nproc)
+else
+    nproc=4
+fi
+
 # capture the output of a command and write it to stout or to a file
 capture() {
     title=$1
@@ -482,6 +488,20 @@ run_nim() {
     check_output "related_posts_nim.json"
 }
 
+run_nim_con() {
+    echo "Running Nim Concurrent" &&
+        cd ./nim_con &&
+        nimble -y install -d &&
+        ./buildopt.sh ${nproc} &&
+        if [ $HYPER == 1 ]; then
+            capture "Nim Concurrent" hyperfine -r $runs -w $warmup --show-output "./build/related_con"
+        else
+            command ${time} -f '%es %Mk' ./build/related_con
+        fi
+
+    check_output "related_posts_nim_con.json"
+}
+
 run_fsharp() {
     echo "Running FSharp (JIT)" &&
         cd ./fsharp &&
@@ -807,6 +827,10 @@ elif [ "$first_arg" = "nim" ]; then
 
     run_nim
 
+elif [ "$first_arg" = "nim_con" ]; then
+
+    run_nim_con
+
 elif [ "$first_arg" = "fsharp" ]; then
 
     run_fsharp
@@ -892,6 +916,7 @@ elif [ "$first_arg" = "all" ]; then
         run_java_graal || echo -e "\n" &&
         run_java_graal_con || echo -e "\n" &&
         run_nim || echo -e "\n" &&
+        run_nim_con || echo -e "\n" &&
         run_fsharp || echo -e "\n" &&
         run_fsharp_con || echo -e "\n" &&
         run_fsharp_con_aot || echo -e "\n" &&
--- a/run.sh
+++ b/run.sh
@@ -494,7 +494,7 @@ run_nim_con() {
         nimble -y install -d &&
         ./buildopt.sh ${nproc} &&
         if [ $HYPER == 1 ]; then
-            capture "Nim Concurrent" hyperfine -r $runs -w $warmup --show-output "./build/related_con"
+            capture "Nim Concurrent (nproc=${nproc})" hyperfine -r $runs -w $warmup --show-output "./build/related_con"
         else
             command ${time} -f '%es %Mk' ./build/related_con
         fi
--- a/nim_con/.gitignore
+++ b/nim_con/.gitignore
@@ -1,0 +1,3 @@
+build/
+default.prof*
+nimcache/
--- a/nim_con/buildopt.sh
+++ b/nim_con/buildopt.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+nproc=${1}
+if [[ -z "${nproc}" ]]; then
+    nproc=4
+fi
+## hardwired
+# nproc=4
+
+chansize=${2}
+if [[ -z "${chansize}" ]]; then
+    chansize=$((2 * ${nproc}))
+fi
+## hardwired
+# chansize=16
+
+profdata="default.profdata"
+profraw="default.profraw"
+
+if [[ ${DANGER} = true ]]; then
+    build_kind=danger
+else
+    build_kind=release
+fi
+
+echo "Compiling profiled executable"
+nim c -d:FixedChanSize=${chansize} \
+      -d:ThreadPoolSize=${nproc} \
+      -d:profileGen \
+      -d:${build_kind} \
+      related_con.nim
+
+echo "Generating profile"
+./build/related_con >/dev/null
+if [[ "$(uname)" = "Darwin" ]]; then
+    xcrun llvm-profdata merge -output="${profdata}" "${profraw}"
+else
+    llvm-profdata merge -output="${profdata}" "${profraw}"
+fi
+
+echo "Compiling optimized executable"
+nim c -d:FixedChanSize=${chansize} \
+      -d:ThreadPoolSize=${nproc} \
+      -d:profileUse \
+      -d:${build_kind} \
+      related_con.nim
--- a/nim_con/buildopt.sh
+++ b/nim_con/buildopt.sh
@@ -9,9 +9,11 @@ fi
 ## hardwired
 # nproc=4
 
+threads=$((2 * ${nproc}))
+
 chansize=${2}
 if [[ -z "${chansize}" ]]; then
-    chansize=$((2 * ${nproc}))
+    chansize=$((2 * ${threads}))
 fi
 ## hardwired
 # chansize=16
@@ -27,7 +29,7 @@ fi
 
 echo "Compiling profiled executable"
 nim c -d:FixedChanSize=${chansize} \
-      -d:ThreadPoolSize=${nproc} \
+      -d:ThreadPoolSize=${threads} \
       -d:profileGen \
       -d:${build_kind} \
       related_con.nim
@@ -42,7 +44,7 @@ fi
 
 echo "Compiling optimized executable"
 nim c -d:FixedChanSize=${chansize} \
-      -d:ThreadPoolSize=${nproc} \
+      -d:ThreadPoolSize=${threads} \
       -d:profileUse \
       -d:${build_kind} \
       related_con.nim
--- a/nim_con/config.nims
+++ b/nim_con/config.nims
@@ -1,0 +1,35 @@
+import std/os
+
+const
+  cfgPath = currentSourcePath.parentDir
+  cacheSubdirHead = joinPath(cfgPath, "nimcache")
+  cacheSubdir = joinPath(cacheSubdirHead,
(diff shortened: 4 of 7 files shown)


michaelsbradleyjr:
cc: @inv2004

jinyus:
error install dep:

```
Verifying dependencies for related_con@0.1.0
     Info:  Dependency on jsony@#head already satisfied
  Verifying dependencies for jsony@1.1.5
 Installing malebolgia@#head
       Tip: 6 messages have been suppressed, use --verbose to show them.
nimble.nim(683)          getDownloadInfo

    Error:  Package malebolgia@#head not found.
```

michaelsbradleyjr:
> error install dep:

Ran fine in my fresh Arch container. Looks similiar to something re: package installation (in general) that people occasionally mention in the Nim discord and forum.

In your shell, if you're okay with removing all currently installed nimble packages, can you try:
```
$ rm -rf "${HOME}/.nimble/"*.json "${HOME}"/.nimble/pkgs*
```
Then rerun:
```
./run.sh nim_con
```

jinyus:
Slower with all payloads on VM too:

old:

https://github.com/jinyus/related_post_gen/blob/b8aa68152b40c8275c0180f7c45435354ff2e2f1/readme.md?plain=1#L72

```
5k:

Processing time (w/o IO): 48.365ms                                                                                                                                                                       
Processing time (w/o IO): 48.044ms                                                                                                                                                                       
Processing time (w/o IO): 47.768ms                                                                                                                                                                       
Processing time (w/o IO): 48.627ms                                                                                                                                                                       
Processing time (w/o IO): 48.039ms                                                                                                                                                                       
Processing time (w/o IO): 48.298ms                                                                                                                                                                       
Processing time (w/o IO): 48.241ms                                                                                                                                                                       
Processing time (w/o IO): 49.092ms                                                                                                                                                                       
Processing time (w/o IO): 48.364ms                                                                                                                                                                       
Processing time (w/o IO): 50.322ms                                                                                                                                                                       
Processing time (w/o IO): 47.901ms                                                                                                                                                                       
Processing time (w/o IO): 47.815ms                                                                                                                                                                       
Processing time (w/o IO): 48.047ms                                                                                                                                                                       
  Time (mean ± σ):      74.9 ms ±   1.0 ms    [User: 145.7 ms, System: 11.4 ms]                                                                                                                          
  Range (min … max):    73.5 ms …  76.4 ms    10 runs

20k:

Processing time (w/o IO): 912.784ms
Processing time (w/o IO): 916.289ms
Processing time (w/o IO): 913.457ms
  Time (mean ± σ):      1.023 s ±  0.001 s    [User: 2.728 s, System: 0.049 s]
  Range (min … max):    1.022 s …  1.023 s    2 runs


60k:

Processing time (w/o IO): 7294.964ms
Processing time (w/o IO): 7285.523ms
Processing time (w/o IO): 7280.419ms
  Time (mean ± σ):      7.659 s ±  0.004 s    [User: 21.528 s, System: 0.094 s]
  Range (min … max):    7.656 s …  7.663 s    2 runs

**Comment by michaelsbradleyjr** [comment]
The VM has 4 cores, right?

**Comment by jinyus** [comment]
Yes. Maybe the scheduler/context-switching overhead is too much when using a small number of cores.
 
I guess it can stay in drafts until it has been improved...or merged but excluded from the charts.

**Comment by michaelsbradleyjr** [comment]
With 16 cores, here's what I'm getting for a 5k run in an Arch container virtualized on top of macOS on a 2019 macbook pro:

```
Nim Concurrent:

Benchmark 1: ./build/related_con
Processing time (w/o IO): 18.857ms
Processing time (w/o IO): 18.168ms
Processing time (w/o IO): 18.472ms
Processing time (w/o IO): 19.205ms
Processing time (w/o IO): 19.911ms
Processing time (w/o IO): 18.213ms
Processing time (w/o IO): 18.606ms
Processing time (w/o IO): 18.917ms
Processing time (w/o IO): 17.665ms
Processing time (w/o IO): 18.462ms
Processing time (w/o IO): 18.982ms
Processing time (w/o IO): 17.335ms
Processing time (w/o IO): 18.913ms
  Time (mean ± σ):      42.3 ms ±   1.1 ms    [User: 178.0 ms, System: 18.0 ms]
  Range (min … max):    40.7 ms …  44.2 ms    10 runs
 
Checking output
Success: related_posts_nim_con.json is valid!
```

Anyway, just trying to be clear that I'm seeing a real speedup for 16 cores/threads (8 too), but there's a bottleneck at 4. I'll be trying to improve the numbers for 4!

**Comment by jinyus** [comment]
This is the score on my PC (8cores/16threads):

```
Processing time (w/o IO): 20.391ms
Processing time (w/o IO): 17.557ms
Processing time (w/o IO): 18.597ms
Processing time (w/o IO): 20.177ms
Processing time (w/o IO): 17.133ms
Processing time (w/o IO): 19.023ms
Processing time (w/o IO): 19.172ms
Processing time (w/o IO): 19.84ms
Processing time (w/o IO): 17.573ms
Processing time (w/o IO): 20.328ms
Processing time (w/o IO): 18.771ms
Processing time (w/o IO): 17.56ms
Processing time (w/o IO): 21.15ms
  Time (mean ± σ):      42.3 ms ±   2.1 ms    [User: 191.4 ms, System: 16.0 ms]
  Range (min … max):    39.4 ms …  45.9 ms    10 runs
```

D for comparison:

```
Processing time (w/o IO): 6.65ms
Processing time (w/o IO): 6.462ms
Processing time (w/o IO): 6.839ms
Processing time (w/o IO): 6.763ms
Processing time (w/o IO): 6.282ms
Processing time (w/o IO): 6.817ms
Processing time (w/o IO): 6.719ms
Processing time (w/o IO): 7.511ms
Processing time (w/o IO): 7.154ms
Processing time (w/o IO): 6.648ms
Processing time (w/o IO): 6.272ms
Processing time (w/o IO): 6.776ms
Processing time (w/o IO): 8.362ms
  Time (mean ± σ):      72.1 ms ±   5.5 ms    [User: 84.1 ms, System: 21.6 ms]
  Range (min … max):    65.6 ms …  82.9 ms    10 runs
```

**Comment by michaelsbradleyjr** [comment]
For fun, what do you get on your machine if you do `DANGER=true ./run.sh nim_con`?

D's numbers are looking good!

In Nim there's a restriction that `ref object` can't be shared across threads, it's not even safe in a single-writer/multi-reader setup, and adding locks doesn't help anything. The restriction has to do with Nim's memory model and you end up having to do messaging passing across threads one way or another, which involves copies and it all starts to add up in terms of cpu time.

Things are better with Nim v2.0.0 and its ARC/ORC memory model (vs. the older default `refc` memory model in pre-2.0). And libraries like Araq's `malebolgia` are supposed to make things easier and promote techniques that scale better. There may be some bug/s in `malebolgia`, I'm not seeing quite what I expect when in do `echo getThreadId()` inside `proc process`. In any case, I'm on the hunt!

There may also be some techniques available in Nim v2 that I haven't tried yet (things I've seen mentioned over time), though they may turn out to be in the unsafe category, we'll see.

**Comment by jinyus** [comment]
danger:

```
Processing time (w/o IO): 14.424ms
Processing time (w/o IO): 9.555ms
Processing time (w/o IO): 9.718999999999999ms
Processing time (w/o IO): 8.023ms
Processing time (w/o IO): 9.760999999999999ms
Processing time (w/o IO): 8.231ms
Processing time (w/o IO): 7.394ms
Processing time (w/o IO): 11.686ms
Processing time (w/o IO): 12.434ms
Processing time (w/o IO): 9.055ms
Processing time (w/o IO): 8.356999999999999ms
Processing time (w/o IO): 7.454ms
Processing time (w/o IO): 6.979ms
  Time (mean ± σ):      31.8 ms ±   3.1 ms    [User: 74.3 ms, System: 15.7 ms]
  Range (min … max):    27.9 ms …  38.5 ms    10 runs
```

**Comment by michaelsbradleyjr** [comment]
Just added a commit: when the number of threads is 2 * cores, it seems to do better than before with a small number (`nproc=4`).

Now, that's on my machine (which has a lot more then 4 cores) and it won't necessarily hold up on a box where 4 is all that's available, i.e. your guess about the scheduler/context-switching overhead may be correct and improving 4 core perf would then require other kinds of changes.

But... maybe give the latest change a run on the benchmarking VM. If it doesn't make a difference, sorry for the time waste.

**Comment by michaelsbradleyjr** [comment]
I tried using an explicit `TicketLock` (from the malebolgia pkg), but the performance is about the same so I didn't push that commit to this PR branch.

If curious, the changes can be seen here: [`nim_con...nim_con_tl`](https://github.com/michaelsbradleyjr/related_post_gen/compare/nim_con...michaelsbradleyjr:related_post_gen:nim_con_tl)

**Comment by michaelsbradleyjr** [comment]
I just pushed some commits to this PR branch.

I spent more time with `TicketLock`, the data stored in `digest`, and the writeable data structure shared/locked across threads (in `proc main` it was named `metaCol`, with my latest changes it's `metas`).

I realized a few things, e.g. the compiler should probably have been complaining about `digest` when it was global. If `tagCol` and `tagMap` are setup like `digest` was before my latest changes, the compiler fails with an error re: gc safety, so it's probably a Nim compiler bug that it allowed `digest` as a global.

With latest changes, `tagMap` and `tagCol` are immutable data structures. It's not possible/safe to share `ref` things across threads, but I can send `ptr` for immutable structures to the worker threads and deref with `[]`. There's an experimental compiler feature named `atomicRef`, but I'd rather not go down that road and it's unnecessary anyway.

Now, the Nim Manual says `addr`, `cast`, `ptr` and `pointer` (untraced references) are unsafe. On the other hand, the way of Nim is that sometimes unsafe things are what you do when there's no viable alternative.

For example, In `nim_con` and `nim` we want to use the 64 bits of information in `XXH3_64bits(x)` (returns a `uint64`) as a std `Hash` (an alias for `int`, which is the same as `int64` on 64-bit platforms). How can we do a lossless transform? With `cast[Hash]()`. The Manual says `cast` is unsafe but for this purpose it's how you do it.

In `nim_con`, there's immutable *non-ref* data on the main thread that we want to access from the worker threads without expending cpu time to copy it to the workers. In Nim v2.0 with ARC/ORC (`--mm:orc` is the Nim compiler's default) that's doable, but how you do it is with `addr`/`ptr`/`[]`. Note that this *does not* disable runtime checks like bounds-checking, etc. Maybe a future version of Nim will be smart enough to better figure out what's immutable so `ptr` etc. don't have to be used manually, but that's not how it works currently.

So that's my case re: why these changes aren't against the rules. Feel free to tell me to get out of town! 😄

With latest changes the performance for `nproc=4` is on par with what I was seeing circa https://github.com/jinyus/related_post_gen/pull/291#issuecomment-1763708299, but with the changes I feel more confident about the code. Also, I cleaned up some extra type names that were helpful when experimenting with changes in multiple funcs/procs at the same time but aren't necessary now.

I have any idea to cut down on allocations re: `top5`, which may improve performance a bit, we'll see.

**Comment by michaelsbradleyjr** [comment]
I refactored `proc top5` and related types, which improves efficiency in several computations (less allocations on each thread so shaves off time).

Also, I had a realization:

The shared data structure the worker threads write to is initialized like this:
```nim
var metas = newSeq[PostMeta](posts.len)
```
Quoting the [Nim Manual](https://nim-lang.org/docs/system.html#newSeq)
> ```
> proc newSeq[T](len = 0.Natural): seq[T]
> ```
> Creates a new sequence of type seq[T] with length len.
> Note that the sequence will be filled with zeroed entries. After the creation of the sequence you should assign entries to the sequence instead of adding them.

So for every index `i in 0..<metas.len` there exists an initialized `PostMeta` object (non-ref) *before* the tasks are spawned on the worker threads.

The way my `nim_con` implementation works, no worker thread will ever write to the same index in `metas`, therefore `TicketLock` is not needed, i.e. the threads can write concurrently to the non-ref data in `var metas`, which results in a nice perf boost.

On my mac laptop I changed Docker Desktop's settings so that containers have a CPU limit of 4, restarted, and here's what I get:

```
$ ./run.sh nim_con
...
Nim Concurrent (nproc=4):

Benchmark 1: ./build/related_con
Processing time (w/o IO): 14.695ms
Processing time (w/o IO): 15.107ms
Processing time (w/o IO): 14.62ms
Processing time (w/o IO): 14.709ms
Processing time (w/o IO): 17.577ms
Processing time (w/o IO): 19.643ms
Processing time (w/o IO): 16.235ms
Processing time (w/o IO): 16.851ms
Processing time (w/o IO): 20.247ms
Processing time (w/o IO): 21.839ms
Processing time (w/o IO): 20.315ms
Processing time (w/o IO): 19.625ms
Processing time (w/o IO): 17.313ms
  Time (mean ± σ):      41.9 ms ±   2.6 ms    [User: 65.7 ms, System: 9.5 ms]
  Range (min … max):    38.7 ms …  46.0 ms    10 runs
 
Checking output
Success: related_posts_nim_con.json is valid!
```

I'm curious if the *4vCPU* in the benchmarking VM's specs (per readme) equates to `nproc=4` or `nproc=8`. For example, my laptop has 8 cores, but the `nproc` command reports 16. It's double because of hyperthreading.

Whether nproc is 4 or 8 on the VM, I'm interested to see how `nim_con` performs on it with the latests changes in this PR branch!

I'm ready to take this PR out of draft, but I'll defer to you.

jinyus:
nproc is 4 on the VM. Still slower that the single threaded version but I will merge and exclude from the charts

run.sh
```diff
@@ -482,6 +488,21 @@ run_nim() {
     check_output "related_posts_nim.json"
 }
 
+run_nim_con() {
+    echo "Running Nim Concurrent" &&
+        cd ./nim_con &&
+        echo "using ${nproc} threads" &&
```
michaelsbradleyjr, line 494: This isn't quite right because `nim_con/buildopt.sh` has
```sh
threads=$((2 * ${nproc}))
```
But no big deal.

michaelsbradleyjr:
> nproc is 4 on the VM. Still slower that the single threaded version but I will merge and exclude from the charts

I'm disappointed to learn it's still slower, because with `nproc=4` in local VM/container I was seeing a nice improvement with latest changes. But I never give up! 🤯 ... though I will take a break from it for awhile. 😄

PR #299, opened 2023-10-16
Naive Haskell solution

schoettl:
to be optimized... by more strictness, better data types, etc.


jinyus:
Hey, it needs to output the processing time:

```ruby
print "Processing time (w/o IO): #{(t2 - t1).total_milliseconds}ms\n"

```

schoettl:
OK. Parsing and encoding JSON does still count as IO, right?

jinyus ***@***.***> schrieb am Di., 17. Okt. 2023, 01:34:

schoettl:
I don't know yet how compute the processing time... Haskell's lazyness does everything inside the `do` block in random order ^^

jinyus:
> OK. Parsing and encoding JSON does still count as IO, right? 

It should exclude json parsing and writing the results to disk.

jinyus:
That behavior is very weird, are you still working on it? I'm not sure I should merge...

schoettl:
Thanks @ozkutuk for the suggestion on how to force evaluation. The timer should now work.

Ready to merge from my side, optimizations can be provided later.

PR #304, opened 2023-10-17
[Csharp] Use frozen dictionaries

MichalStrehovsky:
* Use `FrozenDictionary` to look up tags.
* Factor out `GetRelatedPost` so that we can use `stackalloc` that the compiler zero inits faster than what `Span.Clear` can do
* Add tight loop to find next highest `taggedPostCount`
* AOT only: [Optimize for speed](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/optimizing).

PR #309, opened 2023-10-17
AOT optimize for speed for F# as in #304

delneg:
https://github.com/jinyus/related_post_gen/pull/304
As per this PR


delneg:
That's interesting, certainly didn't expect that

jinyus:
Your new PR is a little faster so I'll remove the speed config and test

PR #325, opened 2023-10-18
[CSharp] Add concurrent implementation.

zigzag312:
C# concurrent implementation based on past PRs from @MichalStrehovsky. 

Biggest gain comes from `Parallel.For`. 

Vectors add very slight performance boost on my machine, compared to only `Parallel.For`.


danmoseley:
@MichalStrehovsky can explain what optimize for speed does under the covers.

MichalStrehovsky:
> @MichalStrehovsky can explain what optimize for speed does under the covers.

Mostly it uses different inlining heuristics. It does not allow using different ISA extension. That would be need to be documented. `<IlcInstructionSet>native</IlcInstructionSet>` is the only way to influence what ISA extensions get used by codegen.

PR #328, opened 2023-10-18
[D] SIMD algo for Concurrent version

cyrusmsk:
Based on SIMD code from C#.
Please check the performance improvement before merging.


cyrusmsk:
Wrong result on 5K. Let me check one more time

PR #333, opened 2023-10-18
echo gcc and clang versions

inv2004:
It is hard to find default `gcc` or `clang` versions on the runners. The PR could help with it a bit


jinyus:
Would this fail if one is not installed?

inv2004:
I do not think, because of pipe

Issue #272, opened 2023-10-15
Can zig use ReleaseFast to build the target ?


kaiwu:
Turns out it out-performs C++ with `ReleaseFast` 
With current `ReleaseSafe` approach, it safe guards everything in *runtime*, which adds significant overhead.
If we squat low enough and examine  the top scorers, none of them is doing this.

jinyus:
Turning off safety is against the rules. If other solutions are doing this, please point them out...as avoiding safety checks that others are doing is not fair.

https://ziglang.org/documentation/master/#toc-ReleaseFast

kaiwu:
none of them are doing it in *runtime* I believe. For example, rust and cpp are checking them only in compile time.
exactly this is not fair for zig

jinyus:
IIUC rust does bounds checks at runtime too: https://nnethercote.github.io/perf-book/bounds-checks.html

and cpp is using `.at` so it also does at runtime. 

Go and d also do runtime checks so allowing zig to remove them isn't fair.

cyrusmsk:
> Rust does not make overflow checks in release, but Nim does.

Can you show docs about that? I’ve tried to find myself, but it wasn’t clearly written.

inv2004:
https://doc.rust-lang.org/cargo/reference/profiles.html#release

cyrusmsk:
> https://doc.rust-lang.org/cargo/reference/profiles.html#release

But it seems (https://rust-lang.github.io/packed_simd/perf-guide/bound_checks.html) that only specifically marked part of the language (unsafe, *_unchecked) is not checked for boundaries in case "debug-assertion=false". The rest of the code in general still checked, in case compiler itself couldn't "prove" that this part could be skipped for checking.

inv2004:
@cyrusmsk 
I am about overflows check, you are about bounds check

cyrusmsk:
> @cyrusmsk
> I am about overflows check, you are about bounds check

Oh. Maybe then add flags from this forum topic? https://github.com/nim-lang/Nim/issues/14504

inv2004:
I think that overflow is dangerous - better to enable it for rust. On the same way - what to do with c/c++?

PR #357, opened 2023-10-23
[Nim] test overflow checks disabled

inv2004:
The PR just to check only.

The reasons for it:
- Rust does not have the check in `release` mode at all
- Cpp does not have it at all

### DIFF
--- a/nim/config.nims
+++ b/nim/config.nims
@@ -3,6 +3,7 @@ switch("cc", "gcc")
 switch("panics", "off")
 switch("passC", "-s -flto")
 switch("passL", "-s -flto")
+switch("overflowChecks", "off")
 
 when defined(profileGen):
   echo "Build with profileGen"


michaelsbradleyjr:
Good thought to check it.

Seeing this PR, I tried the same with `nim_con` and it made it slower.

For `nim` it didn't seem to make a difference. Numbers from Azure VM
```
Nim:

	Benchmark 1: ./related
	Processing time (w/o IO): 25.63ms
	Processing time (w/o IO): 25.382ms
	Processing time (w/o IO): 25.599ms
	Processing time (w/o IO): 25.603ms
	Processing time (w/o IO): 25.507ms
	Processing time (w/o IO): 25.613ms
	Processing time (w/o IO): 25.741ms
	Processing time (w/o IO): 25.496ms
	Processing time (w/o IO): 25.401ms
	Processing time (w/o IO): 25.404ms
	Processing time (w/o IO): 25.394ms
	Processing time (w/o IO): 25.432ms
	Processing time (w/o IO): 25.59ms
	 Time (mean ± σ):      52.9 ms ±   0.6 ms    [User: 41.1 ms, System: 11.9 ms]
	 Range (min … max):    52.2 ms …  53.9 ms    10 runs
	
Nim:

	Benchmark 1: ./related
	Processing time (w/o IO): 352.201ms
	Processing time (w/o IO): 348.941ms
	Processing time (w/o IO): 348.982ms
	 Time (mean ± σ):     456.8 ms ±   1.1 ms    [User: 425.4 ms, System: 31.2 ms]
	 Range (min … max):   456.0 ms … 457.6 ms    2 runs
	
Nim:

	Benchmark 1: ./related
	Processing time (w/o IO): 3070.441ms
	Processing time (w/o IO): 3063.702ms
	Processing time (w/o IO): 3063.212ms
	 Time (mean ± σ):      3.405 s ±  0.012 s    [User: 3.295 s, System: 0.105 s]
	 Range (min … max):    3.396 s …  3.413 s    2 runs
```

michaelsbradleyjr:
I made an interesting "discovery" earlier. If you put `switch("threads", "off")` in `nim/config.nims` it causes a significant slowdown.

Given the program is single-threaded, that surprised me quite a bit! But maybe it's a known thing/bug?

inv2004:
@michaelsbradleyjr it is very possible - some code layouts just does not goes right. The same could be applicable to the overflowchecks. I agree - the change could improve it only a bit and _if it works_ - locally it works, but in the VM - not sure

Anyway, I am pretty sure the better result would be in clang+PGO, but still not sure why it does not work

jinyus:
Addressed here: https://github.com/jinyus/related_post_gen/issues/272#issuecomment-1763416408

inv2004:
Did not get it. Because it is about overflows, not bounds

PR #365, opened 2023-10-24
[Nim] Use ref to fill related

inv2004:
@jinyus 
If I can ask to bench it before commit. Locally it shows about -1ms

Thank you

### DIFF
--- a/nim/src/related.nim
+++ b/nim/src/related.nim
@@ -21,17 +21,13 @@ const
 func hash(x: string): Hash =
   cast[Hash](XXH3_64bits(x))
 
-func findTop5(posts: seq[Post], taggedPostCount: seq[uint8]): seq[Post] =
-  result = newSeq[Post](5)
-  var top5: array[5, tuple[idx: int, count: uint8]]
+func findTop5(posts: seq[Post], taggedPostCount: seq[uint8]): array[5, tuple[idx: int, count: uint8]] =
   for i, count in taggedPostCount:
-    if count > top5[4].count:
-      top5[4] = (idx: i, count: count)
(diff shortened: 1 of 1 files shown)


jinyus:
Not merging since it makes the code less concise with no perf benefits.

michaelsbradleyjr:
@inv2004 in case it interests you and you'd like to experiment with it:

I tried to get creative and made some changes in my fork to the single-threaded `nim` so that `posts` seq is mutated directly instead of having another `seq[RelatedPosts]`:

[`michaelsbradleyjr:related_post_gen:main...michaelsbradleyjr:related_post_gen:nim_rev_10`](https://github.com/michaelsbradleyjr/related_post_gen/compare/main...michaelsbradleyjr:related_post_gen:nim_rev_10)

It's just a bit faster for the 5k runs, but then slower than current `main` branch times for 20k and 60k runs.

I tried quite a few variations for that approach, but that's the fastest I could make it. I experimented with `var` mutation vs. return for `findTop5`, `seq` vs. `array` for `related:`, etc., etc. Note: I've turned on `mm:arc` aware that it doesn't deal with cycles and my `nim_rev_10` does involve cycles, but in this case it doesn't leak because of the way `posts` is self-contained.

`clang`, plus the inlining hints, plus `mm:arc`, *minus* xxhash gives the best times. But maybe you can think of a way to tweak it further.

If you want, you can hit me up in the Nim discord and I can get you access to my free-tier Azure VM that has the same setup as the official benchmarking VM. I'll run out of free-tier credits in less than a month and plan to delete it at that time, so might as well get the most out of it.

PR #378, opened 2023-10-28
Slight simplification of Related.jl

Tortar:
Shouldn't affect runtime (at least it doesn't on my machine), just a bit more idiomatic imo


Moelf:
>at least it doesn't on my machine

Haha, I've learned not to trust my local machine in this process

PR #379, opened 2023-10-28
Use one(T), zero(T) instead of 1, 0 in Julia implementation

Tortar:
added one(T) and zero(T) instead of 1 and 0, it was already done inside one of the loops where it counts more for performance, here I added it more for consistency, but it should be (almost **un**noticeably) faster

### DIFF
--- a/julia/Related/src/Related.jl
+++ b/julia/Related/src/Related.jl
@@ -20,7 +20,7 @@ end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
 
-function fastmaxindex!(xs::Vector, topn, maxv)
+function fastmaxindex!(xs::Vector, maxn, maxv)
     maxn .= 1
     maxv .= 0
     top = maxv[1]
--- a/julia/Related/src/Related.jl
+++ b/julia/Related/src/Related.jl
@@ -20,7 +20,7 @@ end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
 
-function fastmaxindex!(xs::Vector, maxn, maxv, ::Type{T}) where {T}
+function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
     maxn .= one(T)
     maxv .= zero(T)
     top = maxv[1]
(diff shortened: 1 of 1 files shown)


Moelf:
That one is actually needed if I remember correctly

julia/Related/src/Related.jl
```diff
@@ -20,9 +20,9 @@ end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
 
-function fastmaxindex!(xs::Vector, topn, maxn, maxv)
-    maxn .= 1
-    maxv .= 0
+function fastmaxindex!(xs::Vector, maxn, maxv, ::Type{T}) where {T}
```
Moelf, line 23: Why do you need an extra argument? Just use eltype of maxn

julia/Related/src/Related.jl
```diff
@@ -45,10 +45,9 @@ end
 function related(posts)
     Ts = (UInt8, UInt16, UInt32, UInt64)
     i = findfirst(T -> length(posts) < typemax(T), Ts)
-    return related(Ts[i], posts)
+    return related(posts, Ts[i])
 end
-function related(::Type{T}, posts) where {T}
-    topn = 5
+function related(posts, ::Type{T}) where {T}
```
Moelf, line 50: What's the point of this?

Moelf:
Our local testing sometimes miss the CI machines performance characteristics, please stop micro adjusting it for style or something else

Moelf:
For example: https://github.com/jinyus/related_post_gen/pull/182

julia/Related/src/Related.jl
Moelf: Or type maxn with `::AbstractVector{T}` and use the T in function body

Moelf (approved):

Tortar:
wow, you are totally right, I rerun the main branch vs. this PR (before last commit) and saw a regression, sorry for that, strangely enough I didn't see it the first time I ran the benchmarks, now I took extra care to benchmark and it should be fine. I think I learnt what you said in the other PR :D

PR #383, opened 2023-10-29
Restore version with StaticArrays for concurrent Julia

Tortar:
@Moelf @MasonProtter 

50% faster on my machine

### DIFF
--- a/julia/Related/src/Related.jl
+++ b/julia/Related/src/Related.jl
@@ -41,9 +41,11 @@ function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
 end
 
 function related(posts)
-    Ts = (UInt8, UInt16, UInt32, UInt64)
-    i = findfirst(T -> length(posts) < typemax(T), Ts)
-    return related(Ts[i], posts)
+    for T in (UInt8, UInt16, UInt32, UInt64)
+        if length(posts) < typemax(T)
+            return related(T, posts)
+        end
+    end
 end
 function related(::Type{T}, posts) where {T}
     topn = 5
--- a/julia/Related/src/Related.jl
+++ b/julia/Related/src/Related.jl
@@ -41,11 +41,9 @@ function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
 end
 
 function related(posts)
-    for T in (UInt8, UInt16, UInt32, UInt64)
-        if length(posts) < typemax(T)
-            return related(T, posts)
-        end
-    end
+    Ts = (UInt8, UInt16, UInt32, UInt64)
+    i = findfirst(T -> length(posts) < typemax(T), Ts)
+    return related(Ts[i], posts)
 end
 function related(::Type{T}, posts) where {T}
     topn = 5
--- a/julia_con/RelatedCon/src/RelatedCon.jl
+++ b/julia_con/RelatedCon/src/RelatedCon.jl
@@ -21,7 +21,7 @@ end
 
 
 function fastmaxindex!(xs::Vector, topn, maxn, maxv)
-    maxn .= 1
+    maxn .= one(UInt32)
     maxv .= zero(UInt32)
     top = maxv[1]
     for (i, x) in enumerate(xs)
@@ -55,7 +55,7 @@ function related(posts)
 
     @threads for (postsrange, _) in chunks(posts, nthreads())
         topn = 5
-        maxn = MVector{topn,Int}(undef)
+        maxn = MVector{topn,UInt32}(undef)
         maxv = MVector{topn,UInt32}(undef)
         taggedpostcount = Vector{UInt32}(undef, length(posts))
 
--- a/julia_con/RelatedCon/src/RelatedCon.jl
+++ b/julia_con/RelatedCon/src/RelatedCon.jl
@@ -1,3 +1,6 @@
+module RelatedCon
+
+using JSON3, StructTypes, Dates, StaticArrays, ChunkSplitters
 using Base.Threads: @threads, nthreads
 
 export main
@@ -18,23 +21,21 @@ end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
 
-function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
-    maxn .= one(T)
-    maxv .= zero(T)
-    top = maxv[1]
+function fastmaxindex!(xs::Vector{T}, topn, maxs) where {T}
+    # each element is a pair idx => val
+    maxs .= (one(T) => zero(T))
+    top = maxs[1][2]
     for (i, x) in enumerate(xs)
         if x > top
-            maxn[1], maxv[1] = i, x
+            maxs[1] = (i => x)
             for j in 2:topn
-                if maxv[j-1] > maxv[j]
-                    maxv[j-1], maxv[j] = maxv[j], maxv[j-1]
-                    maxn[j-1], maxn[j] = maxn[j], maxn[j-1]
+                if maxs[j-1][2] > maxs[j][2]
+                    maxs[j-1], maxs[j] = maxs[j], maxs[j-1]
                 end
             end
-            top = maxv[1]
+            top = maxs[1][2]
         end
     end
-    reverse!(maxn)
     return
 end
 
@@ -58,8 +59,7 @@ function related(::Type{T}, posts) where {T}
 
     @threads for (postsrange, _) in chunks(posts, nthreads())
         topn = 5
-        maxn = MVector{topn,T}(undef)
-        maxv = MVector{topn,T}(undef)
+        maxs = MVector{topn, Pair{Int, T}}(undef)
         taggedpostcount = Vector{T}(undef, length(posts))
 
         for i in postsrange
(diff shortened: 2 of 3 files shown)


julia_con/RelatedCon/src/RelatedCon.jl
```diff
@@ -1,33 +1,29 @@
...
+    related::SVector{topn, PostData}
 end
 
 StructTypes.StructType(::Type{PostData}) = StructTypes.Struct()
 
-function maxindex!(xs::Vector, maxs)
+function fastmaxindex!(xs::Vector{T}, topn, maxs) where {T}
     # each element is a pair idx => val
-    maxs .= (1 => 0)
+    maxs .= (one(T) => zero(T))
```
Moelf, line 26: is there any reason we use this `::Pair` implementation instead of two `MVector` like the non-concurrent version?

Tortar: I'm not sure actually, I think @MasonProtter changed it, it seems to me they require less indexing operations in fastmaxindex, I assumed that if it was changed it should be better. On my machine they look identical performance wise

MasonProtter: Either way, the first element should always be an `Int`, not a `one(T)` since it's an index to an array.

julia_con/RelatedCon/src/RelatedCon.jl
```diff
@@ -40,43 +36,52 @@ function maxindex!(xs::Vector, maxs)
...
     end
-    maxs
+    return
 end
 
-function related!(posts, tagmap, relatedposts)
+function related(posts)
+    Ts = (UInt8, UInt16, UInt32, UInt64)
+    i = findfirst(T -> length(posts) < typemax(T), Ts)
+    return related(Ts[i], posts)
```
MasonProtter, line 45: Is this really a good idea? I haven't checked but I'd worry that the compiler wouldn't like this from a type stability POV. The old way this was done was to do
```
for T in (UInt8, UInt16, UInt32, UInt64)
    if length(posts) < typemax(T)
        related(T, posts)
    end
end
```
which should be a little easier for the compiler to reason about

Tortar: let's use this way then in both versions. Let's see risk averse :D

julia_con/RelatedCon/src/RelatedCon.jl
Tortar: is it okay to return to previous methodology @MasonProtter ? I think it doesn't change much but for consistency we could use one or the other in both versions

Tortar: if you have the time you can maybe look into it to see if it does change something on your macines

Tortar: upon further testing, I see that it is maybe a bit faster the other way so I restored that, hopefully we all agree that it is okay to use it

julia_con/RelatedCon/src/RelatedCon.jl
Moelf:
```julia
julia> function f_long(posts)
           for T in (UInt8, UInt16, UInt32, UInt64)
               if length(posts) < typemax(T)
                   return typemin(T)
               end
           end
       end
f_long (generic function with 1 method)

julia> function f_short(posts)
           Ts = (UInt8, UInt16, UInt32, UInt64)
           i = findfirst(T -> length(posts) < typemax(T), Ts)
           return typemin(Ts[i])
       end
f_short (generic function with 1 method)

julia> @benchmark f_short(a) setup=begin a = rand(rand([10, 100, 5000, 50000, 10^6])) end
BenchmarkTools.Trial: 10000 samples with 962 evaluations.
 Range (min … max):  85.252 ns … 138.394 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     87.793 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   90.372 ns ±   6.200 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁▅▇▇▅▃▄█▇▃                                    ▁▁▄▅▆▅▃▂       ▂
  ██████████▅▆▆▇█▇█▇█▇▆▇█▇█▇▆▇██▆▆▄▅▃▅▄▄▅▃▄▅▄▃▁▅█████████▆▅▅▃▅ █
  85.3 ns       Histogram: log(frequency) by time       105 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark f_long(a) setup=begin a = rand(rand([10, 100, 5000, 50000, 10^6])) end
BenchmarkTools.Trial: 10000 samples with 585 evaluations.
 Range (min … max):  202.494 ns …   1.774 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     325.138 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   316.423 ns ± 111.901 ns  ┊ GC (mean ± σ):  0.07% ± 1.06%

  █▃▂▂                 ▇▆▃▁▁▂                      ▆▄▂          ▂
  ████▇▆▄▄▄▄▄▃▄▃▂▆▄▅▅▂▃██████▇▇▇▅▆▆▄▂▅▂▄▅▅▅▅▃▄▃▄▄▃▄████▇▇▇█▇▅▅▅ █
  202 ns        Histogram: log(frequency) by time        547 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

this is very funny

Tortar: I know that! I microbenchmarked that too before suggesting it

Moelf: yeah computers are hard LOL, let's just keep the loop anyway

tecosaur: What's the codegen difference?

Moelf: there are somehow more boxing for long:
```julia
julia> @code_warntype f_long([1,2,3])
MethodInstance for f_long(::Vector{Int64})
  from f_long(posts) @ Main REPL[10]:1
Arguments
  #self#::Core.Const(f_long)
  posts::Vector{Int64}
Locals
  @_3::Union{Nothing, Tuple{DataType, Int64}}
  T::DataType
Body::Any
1 ─ %1  = Core.tuple(Main.UInt8, Main.UInt16, Main.UInt32, Main.UInt64)::Core.Const((UInt8, UInt16, UInt32, UInt64))
│         (@_3 = Base.iterate(%1))
│   %3  = (@_3::Core.Const((UInt8, 2)) === nothing)::Core.Const(false)
│   %4  = Base.not_int(%3)::Core.Const(true)
└──       goto #6 if not %4
2 ┄ %6  = @_3::Tuple{DataType, Int64}
│         (T = Core.getfield(%6, 1))
│   %8  = Core.getfield(%6, 2)::Int64
│   %9  = Main.length(posts)::Int64
│   %10 = Main.typemax(T)::Any
│   %11 = (%9 < %10)::Any
└──       goto #4 if not %11
3 ─ %13 = Main.typemin(T)::Any
└──       return %13
4 ─       (@_3 = Base.iterate(%1, %8))
│   %16 = (@_3 === nothing)::Bool
│   %17 = Base.not_int(%16)::Bool
└──       goto #6 if not %17
5 ─       goto #2
6 ┄       return nothing

julia> @code_warntype f_short([1,2,3])
MethodInstance for f_short(::Vector{Int64})
  from f_short(posts) @ Main REPL[11]:1
Arguments
  #self#::Core.Const(f_short)
  posts::Vector{Int64}
Locals
  #7::var"#7#8"{Vector{Int64}}
  i::Union{Nothing, Int64}
  Ts::NTuple{4, DataType}
Body::Any
1 ─      (Ts = Core.tuple(Main.UInt8, Main.UInt16, Main.UInt32, Main.UInt64))
│   %2 = Main.:(var"#7#8")::Core.Const(var"#7#8")
│   %3 = Core.typeof(posts)::Core.Const(Vector{Int64})
│   %4 = Core.apply_type(%2, %3)::Core.Const(var"#7#8"{Vector{Int64}})
│        (#7 = %new(%4, posts))
│   %6 = #7::var"#7#8"{Vector{Int64}}
│        (i = Main.findfirst(%6, Ts::Core.Const((UInt8, UInt16, UInt32, UInt64))))
│   %8 = Base.getindex(Ts::Core.Const((UInt8, UInt16, UInt32, UInt64)), i)::DataType
│   %9 = Main.typemin(%8)::Any
└──      return %9
```

julia/Related/src/Related.jl
```diff
@@ -41,9 +41,11 @@ function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
 end
 
 function related(posts)
-    Ts = (UInt8, UInt16, UInt32, UInt64)
-    i = findfirst(T -> length(posts) < typemax(T), Ts)
-    return related(Ts[i], posts)
+    for T in (UInt8, UInt16, UInt32, UInt64)
+        if length(posts) < typemax(T)
```
jinyus, line 45: This will no longer be allowed. Use a static size that abides by the rules. ie: uint32.

Moelf: we need to also make a change to the non-concurrent version I guess.

Tortar: did it for the concurrent version

Tortar:
I think it can be merged

PR #385, opened 2023-10-29
[Julia] Always use UInt32

Moelf:


julia/Related/src/Related.jl
```diff
@@ -89,7 +85,7 @@ end
 function main()
     json_string = read(@__DIR__()*"/../../../posts.json", String)
     posts = JSON3.read(json_string, Vector{PostData})
-    fake_posts = fill(posts[1], length(posts))
+    fake_posts = posts[1:2]
```
Moelf, line 88: since we don't look-up according to length at runtime anymore, we don't need `fake_posts` to have the same length

PR #392, opened 2023-10-31
daScript initial implementation

zigzag312:
This is my first time using this language. I couldn't figure it out how to avoid cloning without using unsafe. So, memory usage may get high with bigger posts count.

### DIFF
--- a/dascript/related.das
+++ b/dascript/related.das
@@ -1,0 +1,87 @@
+require daslib/json
+require fio
+require daslib/json_boost
+require strings
+
+struct Post
+    _id: string
+    title: string
+    tags: array<string>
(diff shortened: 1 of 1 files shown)


jinyus:
How do you get the compiler? I'm not seeing any download links or build instructions?

zigzag312:
https://github.com/GaijinEntertainment/daScript/blob/master/doc/getting_started.md

On Windows I updated `generate_msvc_2019.bat` for VS 2022, to generate the `.sln` and build it with VS2022. 

I used 0.4 branch as it the latest stable version. 

I couldn't get VSCode language server to work though.

PR #394, opened 2023-11-01
reorder to move parts into functions, switch to c++20 and gcc

hassec:
- Moved the various parts into functions to show that one doesn't have to use a single large main function.
- some small performance improvements int top5 logic by improving memory locality.
- moved to gcc & c++20 since that was the fastest after changes.

@jinyus would be curious to see how this performs on the VM.

(Tried adding better maps but only got better results when using `march=x86_64_v3` but seems like that is forbidden here from what I understand)

### DIFF
--- a/cpp/main.cpp
+++ b/cpp/main.cpp
@@ -1,20 +1,16 @@
 #include <iostream>
 #include <vector>
 #include <string>
-#include <unordered_map>
 #include <fstream>
 #include <chrono>
-#include <algorithm>
-#include <array>
 #include <stdint.h>
+#include <unordered_map>
 #include "include/json.hpp" // Assuming this path is correct
 
-// NEEDS IMPROVEMENT: Excluded from charts until then
-
 using json = nlohmann::json;
-
-const size_t INITIAL_TAGGED_COUNT_SIZE = 100;
-const size_t TOP_N = 5;
+using map_t = std::unordered_map<std::string, std::vector<int>>;
+constexpr size_t INITIAL_TAGGED_COUNT_SIZE = 100;
+constexpr size_t TOPN=5;
 
 struct Post
 {
@@ -27,7 +23,7 @@ struct RelatedPosts
 {
     std::string _id;
     std::vector<std::string> tags;
-    std::vector<Post *> related;
+    std::vector<Post const*> related;
 };
 
 void to_json(json &j, const RelatedPosts &rp)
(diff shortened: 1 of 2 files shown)


cpp/main.cpp
```diff
@@ -107,11 +111,12 @@ int main()
...
         taggedPostCount[i] = 0;
 
-        uint8_t top5[5] = {0, 0, 0, 0, 0};
+        uint8_t top5[TOPN] = {0, 0, 0, 0, 0};
+        int related[TOPN] = {0, 0, 0, 0, 0};
         uint8_t minTags = 0;
 
         //  custom priority queue to find top N
-        for (int j = 0; j < taggedPostCount.size(); j++)
+        for (size_t j = 0; j < total; j++)
```
Moelf, line 119: isn't this bound-check evading?

cpp/main.cpp
```diff
@@ -45,13 +41,30 @@ void to_json(json &j, const RelatedPosts &rp)
...
-int main()
-{
+
+
+map_t get_tagMap(std::vector<Post> const& posts){
+    map_t tagMap;
+    tagMap.reserve(INITIAL_TAGGED_COUNT_SIZE);
+
+    int total = static_cast<int>(posts.size());
+    for (int i = 0; i < total; ++i)
```
Moelf, line 51: isn't this bound-check evading?

cpp/main.cpp
cyrusmsk: How so? In the logic of the code we know the length of the posts file after reading no it is the same length for this array.

jinyus:
Main:
`C++ | 24.92 ms | 339.00 ms | 2.92 s`

This Pr:
```
C++:

        Benchmark 1: ./main
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 24 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
        Processing time (w/o IO): 23 ms
          Time (mean ± σ):     226.8 ms ±   2.4 ms    [User: 194.4 ms, System: 31.6 ms]
          Range (min … max):   222.4 ms … 230.1 ms    10 runs
         
C++:

        Benchmark 1: ./main
        Processing time (w/o IO): 328 ms
        Processing time (w/o IO): 323 ms
        Processing time (w/o IO): 322 ms
          Time (mean ± σ):      1.148 s ±  0.002 s    [User: 1.008 s, System: 0.133 s]
          Range (min … max):    1.146 s …  1.150 s    2 runs
         
C++:

        Benchmark 1: ./main
        Processing time (w/o IO): 2789 ms
        Processing time (w/o IO): 2781 ms
        Processing time (w/o IO): 2795 ms
          Time (mean ± σ):      5.415 s ±  0.112 s    [User: 4.940 s, System: 0.334 s]
          Range (min … max):    5.336 s …  5.493 s    2 runs

```

cpp/main.cpp
jinyus: storing the size of the input at runtime is allowed.

PR #397, opened 2023-11-01
[Julia] Use inbounds since length is calculated at runtime

Moelf:
according to https://github.com/jinyus/related_post_gen/pull/394#discussion_r1378827340

### DIFF
--- a/julia/Related/src/Related.jl
+++ b/julia/Related/src/Related.jl
@@ -41,6 +41,7 @@ function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
 end
 
 function related(posts)
+    L = length(posts)
     T = UInt32
     topn = 5
     # key is every possible "tag" used in all posts
@@ -54,8 +55,8 @@ function related(posts)
         end
     end
 
-    relatedposts = Vector{RelatedPost}(undef, length(posts))
-    taggedpostcount = Vector{T}(undef, length(posts))
+    relatedposts = Vector{RelatedPost}(undef, L)
+    taggedpostcount = Vector{T}(undef, L)
 
     maxn = MVector{topn,T}(undef)
     maxv = MVector{topn,T}(undef)
@@ -67,7 +68,8 @@ function related(posts)
         # give all related post +1 in `taggedpostcount` shadow vector
         for tag in post.tags
             for idx in tagmap[tag]
-                taggedpostcount[idx] += one(T)
+                # all length are bounded by L = length(posts) we know at runtime
+                @inbounds taggedpostcount[idx] += one(T)
             end
         end
 
--- a/julia/Related/src/Related.jl
+++ b/julia/Related/src/Related.jl
@@ -48,11 +48,13 @@ function related(posts)
     # value is indicies of all "post"s that used this tag
     tagmap = Dict{String,Vector{T}}()
     sizehint!(tagmap, 100)
-    @inbounds for idx in 1:L
+    idx = 1
+    @inbounds while idx <= L
         for tag in posts[idx].tags
             tags = get!(() -> T[], tagmap, tag)
             push!(tags, idx)
         end
+        idx += 1
     end
 
     relatedposts = Vector{RelatedPost}(undef, L)
@@ -61,7 +63,8 @@ function related(posts)
     maxn = MVector{topn,T}(undef)
     maxv = MVector{topn,T}(undef)
 
-    @inbounds for i in 1:L
+    i = 1
+    @inbounds while i<=L
         post = posts[i]
         taggedpostcount .= zero(T)
         # for each post (`i`-th)
@@ -69,6 +72,7 @@ function related(posts)
         # give all related post +1 in `taggedpostcount` shadow vector
         for tag in post.tags
             for idx in tagmap[tag]
+                @assert idx <= L
                 # all length are bounded by L = length(posts) we know at runtime
                 taggedpostcount[idx] += one(T)
             end
@@ -81,6 +85,7 @@ function related(posts)
 
         relatedpost = RelatedPost(post._id, post.tags, SVector{topn}(@view posts[maxn]))
         relatedposts[i] = relatedpost
+        i+=1
     end
 
     return relatedposts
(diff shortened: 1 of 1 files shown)


julia/Related/src/Related.jl
```diff
@@ -41,6 +41,7 @@ function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
 end
 
 function related(posts)
+    L = length(posts)
```
Moelf, line 44: everything are bounded by this `L` so we know everything will be inbound.

jinyus:
This is not allowed and not synonymous to the linked PR. Evading bounds check and turning off bounds checking are different.

Moelf:
how is this different from C++? If you want to enforce bound check, you should require C++ to do
```cpp
for (int j = 0; j < taggedPostCount.size(); j++)
```
no?

Moelf:
>This is not allowed and not synonymous to the linked PR. Evading bounds check and turning off bounds checking are different.

You can't say C++ is not evading because C++ inherently doesn't have bound check, that sounds like an unfair comparison.

Moelf:
effectively every single line in C++ that looks like
```
tags[i]
```

translates to Julia
```
@inbounds tags[i]
```

The question is do you want runtime bound check or not?

cyrusmsk:
I also think that C++ code should use "at(k)" instead of "[k]" as much as possible. I can prepare PR for that.

Regarding "len = length(array)" - I believe compiler with -O3 optimize it anyway, even if we are not going to use it directly. But probably not. So it depends on Rules.

This PR without @inbunds looks reasonable for me. With @inbounds - not.

jinyus:
> how is this different from C++? If you want to enforce bound check, you should require C++ to do
> 
> ```c++
> for (int j = 0; j < taggedPostCount.size(); j++)
> ```
> 
> no?
> 
> If you can pull the `size()` out and say "it's not gonna change at runtime because I know", then you should agree this Julia code is doing the same thing, because everything is tied to the run-time length `L`

I will accept:
```
i = 1
while i <= L
    # code block
    i += 1
end
```

as an equivalent to the c-style for loop with a cached length. 

Explicitly turning off runtime checks are not allowed. So `@inbounds taggedpostcount[idx] += one(T)` is not allowed.

jinyus:
> I also think that C++ code should use "at(k)" instead of "[k]" as much as possible. I can prepare PR for that.

I'm not opposed to that.

Moelf:
@jinyus is this better? I can tell you it's the same but if this style is more of your jam I'm fine with it

jinyus:
Remove  `@inbounds` and we're good

Moelf:
>This PR without @ inbounds looks reasonable for me. With @ inbounds - not.

the `inbounds` just make Julia has the save default indexing behavior as C++, I don't understand what's the problem. Basically here's the analogy:

when you do `ary[i]` in Julia, it actually does a bound check schematically it looks like like:
```
firstindex(ary) <= i <= lastindex(ary) # check this first
@inbounds ary[i] # now access without bounds check otherwise you will StackOverflow in checking bound logic
```

in C++, you just by default always doing `@inbounds ary[i]`, so to me if this C++ code is legal:
```cpp
        for (size_t i{0}; i<TOPN; ++i){
          relatedPost.related[i] = &posts[related[i]];
        }
```

then why is this not legal in Julia?
```julia
i = 0
 while i <= TOPN
    @inbounds relatedPost.related[i] =...
    i+=1
end
```
please explain to me what's the different @jinyus

Moelf:
>Remove @inbounds and we're good

this PR without `@inbounds` is useless, the whole point is that you indicated indexing without runtime bounds checking every time is legal as long as we use runtime calculated length, so I implemented the same logic as C++, i.e. use the same runtime computed length variable for everything and don't do bound checking every single time.

julia/Related/src/Related.jl
```diff
@@ -41,32 +41,39 @@ function fastmaxindex!(xs::Vector{T}, topn, maxn, maxv) where {T}
...
+    i = 1
+    @inbounds while i<=L
+        post = posts[i]
         taggedpostcount .= zero(T)
         # for each post (`i`-th)
         # and every tag used in the `i`-th post
         # give all related post +1 in `taggedpostcount` shadow vector
         for tag in post.tags
             for idx in tagmap[tag]
+                @assert idx <= L
```
Moelf, line 75: notice this btw

jinyus:
If c++ is breaking the rules, then the solution is to fix it, not break the rules too. Your original complaint was about caching the length of the posts which I said is allowed.

Moelf:
I'm very hesitant to change the C++ code base because I'm not a very good C++ programmer.

The point is that in C++ if you cache that variable, you're effectively skipping all the bounds checks by not re-checking the length of the array each iteration

jinyus:
> I'm very hesitant to change the C++ code base because I'm not a very good C++ programmer.
> 
> The point is that in C++ if you cache that variable, you're effectively skipping all the bounds checks by not re-checking the length of the array each iteration

I don't see the correlation with storing the length and indexing into an array.

Moelf:
in this example
```cpp
        for (size_t i{0}; i<TOPN; ++i){
          relatedPost.related[i] = &posts[related[i]];
        }
```

if you don't check `0 < i < relatedPost.size()` every iteration, you're skipping bound check, can you agree?

Moelf:
in this loop for example
https://github.com/jinyus/related_post_gen/blob/668e557cee5c01c5bb42233837d9e7035d19e373/cpp/main.cpp#L51-L55

can you agree not checking `0 <= i && i < posts.size()` every indexing is skipping bound check?

cyrusmsk:
> in this loop for example
> 
> https://github.com/jinyus/related_post_gen/blob/668e557cee5c01c5bb42233837d9e7035d19e373/cpp/main.cpp#L51-L55
> 
> can you agree not checking `0 <= i && i < posts.size()` every indexing is skipping bound check?

C++ version will be fixed.

Moelf:
I mean I almost want to say if you want C++ to do dynamic indexing check you should compile with `-fsanitize=address` but I think performance would be horrible so maybe C++ should just manually sprinkle checks everywhere?

PR #399, opened 2023-11-01
[CPP] Add bounds checking

cyrusmsk:
Changed all static arrays with std::array objects. Also changed all [i] -> at(i) (to use bounds checking)

In this line 'relatedPost = {p._id, p.tags, {0,0,0,0,0}};' I've added explicit initialization with zeros. When I tried to put there 'std::array<Posts const*, TOPN>' - it showed me error.
Initially there were used std::vector{TOPN}.
But it seems vector is more suitable for dynamic arrays, while std::array for fixed sized (https://en.cppreference.com/w/cpp/container/array).

I've tested only on M1 clang++ : the performance became slower. Maybe with g++ it will be better.

Review and suggestions from C++-guys will be appreciated @kaiwu @jinyus @Moelf

### DIFF
--- a/cpp/main.cpp
+++ b/cpp/main.cpp
@@ -1,6 +1,7 @@
 #include <iostream>
 #include <vector>
 #include <string>
+#include <array>
 #include <fstream>
 #include <chrono>
 #include <stdint.h>
@@ -23,7 +24,7 @@ struct RelatedPosts
 {
     std::string _id;
     std::vector<std::string> tags;
-    std::vector<Post const*> related;
+    std::array<Post const*, TOPN> related;
 };
 
 void to_json(json &j, const RelatedPosts &rp)
@@ -50,7 +51,7 @@ map_t get_tagMap(std::vector<Post> const& posts){
     int total = static_cast<int>(posts.size());
     for (int i = 0; i < total; ++i)
     {
-        for (auto const&tag : posts[i].tags)
+        for (auto const&tag : posts.at(i).tags)
         {
             tagMap[tag].emplace_back(i);
         }
@@ -96,9 +97,9 @@ std::vector<RelatedPosts> do_work(std::vector<Post>const& posts,
     for (size_t i = 0; i < total; ++i)
     {
         std::memset(taggedPostCount.data(), 0, total);
-        const Post& p = posts[i];
-        RelatedPosts& relatedPost = allRelatedPosts[i];
-        relatedPost = {p._id, p.tags, std::vector<Post const*>{TOPN}};
+        const Post& p = posts.at(i);
+        RelatedPosts& relatedPost = allRelatedPosts.at(i);
+        relatedPost = {p._id, p.tags, {0,0,0,0,0}};
 
         for (const auto &tag : p.tags)
         {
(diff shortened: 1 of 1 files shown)


Moelf:
I see 17,18 -> 19 ms on my machine which isn't too bad honestly

Moelf (approved):

Moelf:
FIY I tried to apply these changes to `cpp_con` and I get out of bound error at runtime ...

Moelf:
@jinyus you should merge this one, this one is fine, I'm saying the `cpp_con` needs something extra

jinyus:
Hardcoded initialization `{0,0,0,0,0}` isn't allowed so I will mark this as draft until someone more familiar with the lang fixes it. I will try LLMs

Moelf:
then we should use `fill`: https://stackoverflow.com/a/70810681/2960695

jinyus:
Much slower, would like to see a version with just the `.at` changes to compare.

Main:
`C++ | 24.92 ms | 339.00 ms | 2.92 s`

This PR:
```
C++:

        Benchmark 1: ./main
        Processing time (w/o IO): 38 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 38 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 38 ms
        Processing time (w/o IO): 38 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 37 ms
        Processing time (w/o IO): 37 ms
          Time (mean ± σ):     235.3 ms ±   2.0 ms    [User: 207.9 ms, System: 27.1 ms]
          Range (min … max):   232.5 ms … 239.0 ms    10 runs
         
C++:

        Benchmark 1: ./main
        Processing time (w/o IO): 555 ms
        Processing time (w/o IO): 562 ms
        Processing time (w/o IO): 558 ms
          Time (mean ± σ):      1.377 s ±  0.002 s    [User: 1.249 s, System: 0.121 s]
          Range (min … max):    1.375 s …  1.378 s    2 runs
         
C++:

        Benchmark 1: ./main
        Processing time (w/o IO): 4918 ms
        Processing time (w/o IO): 4917 ms
        Processing time (w/o IO): 4909 ms
          Time (mean ± σ):      7.551 s ±  0.068 s    [User: 7.027 s, System: 0.346 s]
          Range (min … max):    7.503 s …  7.600 s    2 runs
```

Moelf:
>Much slower

FWIW, that's largely real because I see a 20% speed-up in Julia by doing `inbounds`

cyrusmsk:
> Much slower, would like to see a version with just the `.at` changes to compare.

but this is the version with just “at”. We can return back the initialization to std::vector. But at my machine there is no any changes in that

jinyus:
It's just weird because locally, it's about 2ms slower.

Moelf:
>It's just weird because locally, it's about 2ms slower.

yeah we all know VM is weird at this point haha

cyrusmsk:
Maybe move results to the local run?
If VM looks so strange?

Moelf:
VM is nice at least it's reproducible (not everyone has the same Intel/AMD/Apple silicon) but in this case because it's vCPU I feel like somehow it's not reproducible in a meaningful way

PR #401, opened 2023-11-01
[daScript] added to run.sh

zigzag312:
@jinyus have you managed to build the compiler?

This PR adds daScript to `run.sh`, but I don't know how to add the compiler.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -746,6 +746,18 @@ run_ruby() {
     check_output "related_posts_ruby.json"
 }
 
+run_dascript() {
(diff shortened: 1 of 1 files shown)


jinyus:
> @jinyus have you managed to build the compiler?
Yes, processing time is ~1.4s on my machine

zigzag312:
I was hoping it would be faster. My implementation is doing way too much cloning. I see even the 5k posts test resulted in OOM. 

I've just asked developers, if there's a way around it without using the unsafe.

PR #415, opened 2023-11-09
Add Gleam

jinyus:
This very is very slow (takes 7s to process 1 post). 
It's very hard to optimize because there're no mutable data structures in the language.


lpil:
Hi there! I'd recommend copying the Erlang one as algorithm used is perfectly suited to Gleam. <https://github.com/jinyus/related_post_gen/blob/main/erlang/src/related_erl.erl>

The algorithm here is one that is intended for mutable imperative languages so it will be catastrophically slow in Gleam.

jinyus:
@lpil man that erlang syntax is giving me an aneurysm, will try though

cyrusmsk:
Gleam is now 1.0

grigio:
+1 for Gleam

jinyus:
@fc1943s This is too slow to merge. I was hoping someone would port the erlang algorithm to gleam. Based on my calculations, it would take more than 9 hours to process 5000 posts.

PR #418, opened 2023-11-11
Add Lobster Implementation

jinyus:
http://aardappel.github.io/lobster/README_FIRST.html

### DIFF
--- a/lobster/related.lobster
+++ b/lobster/related.lobster
@@ -5,7 +5,7 @@ import json
 
 let TOPN = 5
 
-struct Post:
+class Post:
     _id: string
     title: string
     tags: [string]
--- a/lobster/related.lobster
+++ b/lobster/related.lobster
@@ -36,7 +36,7 @@ def main():
                 indexes.push(i)
 
     let posts_count = posts.length
-    var all_related_posts = []::RelatedPost.vector_capacity(posts_count)
+    let all_related_posts = []::RelatedPost.vector_capacity(posts_count)
 
     for(posts) post,i:
 
@@ -71,7 +71,7 @@ def main():
             let index = topn[j*2+1]
             posts[index]
 
-        all_related_posts = all_related_posts.push(RelatedPost {post._id, post.tags, top_posts}) 
+        all_related_posts.push(RelatedPost {post._id, post.tags, top_posts}) 
 
 
     let end = seconds_elapsed()
--- a/lobster/related.lobster
+++ b/lobster/related.lobster
@@ -15,25 +15,6 @@ struct RelatedPost:
     tags: [string]
     related: [Post]
 
-// struct PostwithCount:
-//     id: int
-//     count: int
-
-
-
-def json_to_posts(json_content: string) -> [Post]:
-    let flexbuffer, err = flexbuffers_json_to_binary(json_content)
-
-    assert(not err)
-
-    let posts, err2 = flexbuffers_binary_to_value(typeof [Post], flexbuffer)
-
-    assert(not err2)
-    assert(posts)
-
-    return posts
-
-
 def main():
     let json_content = read_file("../posts.json")
 
@@ -56,33 +37,22 @@ def main():
             else:
                 tag_map.set(tag, push(indexes, i))
 
-    let all_related_posts = []::[RelatedPost]
-    let tagged_post_count = []::[int]
+    var all_related_posts = []::RelatedPost
     let posts_count = posts.length
 
-    let topn_temp = []::[int]
-    let top_posts_temp = []::[Post]
-
     vector_capacity(all_related_posts, posts_count)
-    vector_capacity(tagged_post_count, posts_count)
-    vector_capacity(topn_temp, TOPN*2)
-    vector_capacity(top_posts_temp, TOPN)
-
-    for(TOPN*2) i:
-        topn_temp[i] = 0
 
     for(posts) post,i:
 
-        for (posts_count) j:
-            tagged_post_count[j] = 0
+        let tagged_post_count = map(posts_count) : 0
 
         for (post.tags) tag:
-            for (tag_map.get(tag)) index:
+            for (assert(tag_map.get(tag))) index:
                 tagged_post_count[index] += 1
         
         tagged_post_count[i] = 0
 
-        let topn = copy(topn_temp)
+        let topn = map(TOPN*2) : 0
         var min_tags = 0
 
         for(tagged_post_count) count, idx:
@@ -101,27 +71,20 @@ def main():
 
             min_tags = topn[TOPN*2-2]
 
-        let top_posts = copy(top_posts_temp)
-
-        for(TOPN) j:
+        let top_posts = map(TOPN) j:
             let index = topn[j*2+1]
-            top_posts[j] = posts[index]
+            posts[index]
 
-        all_related_posts[i] = RelatedPost(post._id, post.tags, top_posts)
-
+        all_related_posts = all_related_posts.push(RelatedPost {post._id, post.tags, top_posts}) 
 
 
     let end = seconds_elapsed()
 
-    print("Time: {end - start} seconds")
+    print("Processing time (w/o IO): {end - start}s")
 
-    let related_json = JSON.encode<[RelatedPost]>(all_related_posts)
+    let related_json = JSON.encode(all_related_posts)
 
     write_file("../related.json", related_json)
-
-
-    // print("1st item in map: " + string(tag_map.get("java")).length)
-
 
 
 main()
--- a/lobster/related.lobster
+++ b/lobster/related.lobster
@@ -24,7 +24,7 @@ def main():
 
     let start = seconds_elapsed()
 
-    let tag_map = dictionary<string, [int]> ( 100 )
+    let tag_map = dictionary<string, [int]> ( 101 )
 
     for (posts) post,i:
         for (post.tags) tag:
@@ -33,7 +33,7 @@ def main():
             if not indexes:
                 tag_map.set(tag, [i])
             else:
-                tag_map.set(tag, push(indexes, i))
+                indexes.push(i)
 
     let posts_count = posts.length
     var all_related_posts = []::RelatedPost.vector_capacity(posts_count)
--- a/lobster/related.lobster
+++ b/lobster/related.lobster
@@ -1,0 +1,127 @@
+
+import std
+import dictionary
+import json
+
+let TOPN = 5
+
+struct Post:
+    title: string
+    _id: string
+    tags: [string]
+
+struct RelatedPost:
+    _id: string
+    tags: [string]
+    related: [Post]
+
+// struct PostwithCount:
+//     id: int
+//     count: int
+
+
+
+def json_to_posts(json_content: string) -> [Post]:
+    let flexbuffer, err = flexbuffers_json_to_binary(json_content)
+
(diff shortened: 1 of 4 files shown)


lobster/related.lobster
```diff
@@ -0,0 +1,88 @@
...
+    let tag_map = dictionary<string, [int]> ( 100 )
+
+    for (posts) post,i:
+        for (post.tags) tag:
+            let indexes = tag_map.get(tag)
+
+            if not indexes:
+                tag_map.set(tag, [i])
+            else:
+                tag_map.set(tag, push(indexes, i))
```
aardappel, line 36: no need to re-set `indexes`

lobster/related.lobster
```diff
@@ -0,0 +1,88 @@
...
+def main():
+    let json_content = read_file("../posts.json")
+
+    assert(json_content != nil)
+
+    let posts = JSON.parse<[Post]>(json_content)
+
+    let start = seconds_elapsed()
+
+    let tag_map = dictionary<string, [int]> ( 100 )
```
aardappel, line 27: how many tags are you typically expecting? also may want a prime number as the size (see dictionary.lobster)

lobster/related.lobster
```diff
@@ -0,0 +1,88 @@
...
+    let posts_count = posts.length
+
+    vector_capacity(all_related_posts, posts_count)
+
+    for(posts) post,i:
+
+        let tagged_post_count = map(posts_count) : 0
+
+        for (post.tags) tag:
+            for (assert(tag_map.get(tag))) index:
```
aardappel, line 48: Unlike other higher level languages that have a highly tuned dictionary type built-in, Lobster does not, and this implemented in Lobster dictionary is not particularly fast.. but may be amortized if this loop iterates a lot

lobster/related.lobster
```diff
@@ -0,0 +1,88 @@
...
+        let topn = map(TOPN*2) : 0
+        var min_tags = 0
+
+        for(tagged_post_count) count, idx:
+            guard count > min_tags
+
+            var upper_bound = (TOPN-2) * 2
+
+            while upper_bound >= 0 and count > topn[upper_bound]:
+                topn[upper_bound+2] = topn[upper_bound]
```
aardappel, line 62: This seems to be the most inner of loops? Have you checked for other languages what percentage of time is spent in this loop? Because you risk your benchmark is actually mostly testing the language's ability to do array indexing..

lobster/related.lobster
```diff
@@ -0,0 +1,88 @@
...
+            topn[insert_pos] = count
+            topn[insert_pos+1] = idx
+
+            min_tags = topn[TOPN*2-2]
+
+        let top_posts = map(TOPN) j:
+            let index = topn[j*2+1]
+            posts[index]
+
+        all_related_posts = all_related_posts.push(RelatedPost {post._id, post.tags, top_posts})
```
aardappel, line 76: `push` in-place, no need to re-assign.

lobster/related.lobster
```diff
@@ -0,0 +1,88 @@
+
+import std
+import dictionary
+import json
+
+let TOPN = 5
+
+struct Post:
```
aardappel, line 8: You made this a `struct`, which should generally be faster than a `class` since it is copied rather than dynamically allocated, but it may actually be worse in `for (posts)` because of the copy.. worth trying both and seeing which works better.

lobster/related.lobster
jinyus: Max 100 for this test.

lobster/related.lobster
jinyus: https://github.com/jinyus/related_post_gen/pull/418/commits/54bf9a5185576a8d49b978e5736fde1417672195

lobster/related.lobster
aardappel: Then I would use a prime number like `251` (or if you want to even sure there's less collisions, `1009`).

lobster/related.lobster
jinyus: Dictionaries aren't needed but I made them mandatory for a good balance of a real-world scenario, but I will try an impl where tags are represented as int and use a [[int]] instead; to see how much difference it makes. 

Didn't do much better in [go](https://github.com/jinyus/related_post_gen/blob/aff200fe70220d5ed9a2cdf2665abdc347c1f75c/go/main.go#L49) though.

run.sh
```diff
@@ -758,6 +758,18 @@ run_dascript() {
     check_output "related_posts_dascript.json"
 }
 
+run_lobster() {
+    echo "Running Lobster" &&
+        cd ./lobster &&
+        if [ $HYPER == 1 ]; then
+            capture "Lobster" hyperfine -r $slow_lang_runs -w $warmup --show-output "lobster related.lobster"
+        else
+            command ${time} -f '%es %Mk' lobster related.lobster
```
aardappel, line 767: This is the default JIT run. You'll want to also time a compiled run.

lobster/related.lobster
jinyus: I didn't do any profiling outside of go, this approach was slower in go so a different algo was used. Rust's binary heap was faster that both.

run.sh
jinyus: That's `lobster --cpp related.lobster` right?

aardappel: Yup.. that generates a .cpp, which will need to be compiled and linked with the rest of Lobster: https://aardappel.github.io/lobster/implementation.html

jinyus: Can't get it to output the cpp file: https://github.com/aardappel/lobster/issues/273

aardappel:
So, what are the results?

jinyus:
@aardappel  Just opened an issue(https://github.com/aardappel/lobster/issues/275) because the c++ version isn't outputting the results file. But JIT results are:

```
Lobster (JIT):

        Benchmark 1: lobster related.lobster
        Processing time (w/o IO): 1.374362s
        Processing time (w/o IO): 1.357002s
        Processing time (w/o IO): 1.368519s
        Processing time (w/o IO): 1.357769s
        Processing time (w/o IO): 1.360122s
        Processing time (w/o IO): 1.365586s
        Processing time (w/o IO): 1.368886s
        Processing time (w/o IO): 1.362331s
        Processing time (w/o IO): 1.368122s
        Processing time (w/o IO): 1.369936s
        Processing time (w/o IO): 1.370132s
        Processing time (w/o IO): 1.368973s
        Processing time (w/o IO): 1.358904s
          Time (mean ± σ):      1.451 s ±  0.006 s    [User: 1.430 s, System: 0.021 s]
          Range (min … max):    1.443 s …  1.457 s    10 runs
         
Lobster (JIT):

        Benchmark 1: lobster related.lobster
        Processing time (w/o IO): 21.338143s
        Processing time (w/o IO): 21.351429s
          Time (abs ≡):        21.678 s               [User: 21.614 s, System: 0.059 s]
         
Lobster (JIT):

        Benchmark 1: lobster related.lobster
        Processing time (w/o IO): 189.900035s
        Processing time (w/o IO): 189.729168s
          Time (abs ≡):        190.737 s               [User: 190.542 s, System: 0.171 s]
```

aardappel:
Hmm, those are terrible results even for the JIT mode, close to Python.. pretty sure Lobster is on average at least an order of magnitude faster than Python. Either there is some problem in the Lobster code that we're overlooking where effectively it is doing more work, some algorithmic difference.

Example of a previous benchmark someone made: https://twitter.com/wvo/status/1275870985611177984

jinyus:
Only glaring issue I see is the unoptimized dictionary.

[This line](https://github.com/jinyus/related_post_gen/blob/75761d0c2ce970b26259ddad535836cacf86ba77/lobster/related.lobster#L44) could also be an issue if it's not optimized to a [memset](https://github.com/jinyus/related_post_gen/blob/75761d0c2ce970b26259ddad535836cacf86ba77/cpp/main.cpp#L96). It was slightly faster than allocating a new list every iteration:
```
for(posts_count) j:
            tagged_post_count[j] = 0
```

jinyus:
@aardappel The C++ results are in. ~7x faster than JIT but still in the bottom half of the table.

https://github.com/jinyus/related_post_gen/blob/5b4aa2a529532d76cae59af2cd53884c636b75bd/readme.md?plain=1#L90

aardappel:
I will check why this is all so slow sometime later :)

aardappel:
Ok, had a quick look at why it is so slow, and the first surprising thing is that for the JIT (at 5K posts) I got `1.31s` (very close to your `1.35s`, I have a 5950X), but for the C++ mode I got `104ms` (a LOT faster than your `189ms`), which I think can't be just due to CPUs. This was with the latest visual studio in release mode. That gets it a lot close to the faster languages.

Beyond that, there's nothing too special happening. It spends 50% of its time in the `for (assert(tag_map.get(tag))) index: tagged_post_count[index] += 1` loop (the dictionary doesn't affect it), 35% in `for(posts_count) j: tagged_post_count[j] = 0` (which most definitely doesn't become a memset), and 15% or so in `for(tagged_post_count) count, idx: guard count > min_tags`.

Tried a micro optimisation: move the `tagged_post_count[j] = 0` into the 2nd `for(tagged_post_count)` loop (and `+= 1` -> `++`) and saw time drop to `1.10s` for JIT and C++ to `96ms` so no big gains there.

So yeah, looks like the JIT mode sucks more than I thought, more than 10x slower than C++ mode. Looks like C++ is going to be required when you care for speed. And even that still needs a fair bit more of optimization.

jinyus:
Did a fresh run and the results are stable. Linux vs windows compiler could be the cause of the difference.

https://github.com/jinyus/related_post_gen/blob/9ef36370e2f35977143e56b9ee903dcf5bc9b127/readme.md?plain=1#L97

```
Lobster (C++):

        Processing time (w/o IO): 0.132631s
        total: 0.18s memory: 28096k
        Processing time (w/o IO): 0.13307s
        total: 0.18s memory: 27968k
        Processing time (w/o IO): 0.134092s
        total: 0.18s memory: 28092k
        Processing time (w/o IO): 0.121073s
        total: 0.17s memory: 27964k
        Processing time (w/o IO): 0.132747s
        total: 0.18s memory: 28096k
        Processing time (w/o IO): 0.133872s
        total: 0.18s memory: 28220k
        Processing time (w/o IO): 0.1346s
        total: 0.18s memory: 28220k
        Processing time (w/o IO): 0.12152s
        total: 0.17s memory: 28096k
        Processing time (w/o IO): 0.1224s
        total: 0.17s memory: 28220k
        Processing time (w/o IO): 0.121131s
        total: 0.17s memory: 28096k

Lobster (C++):

        Processing time (w/o IO): 1.99688s
        total: 2.23s memory: 80912k
        Processing time (w/o IO): 2.105853s
        total: 2.34s memory: 80916k
        Processing time (w/o IO): 1.863491s
        total: 2.10s memory: 81044k

Lobster (C++):

        Processing time (w/o IO): 17.828839s
        total: 18.58s memory: 249092k
        Processing time (w/o IO): 17.919368s
        total: 18.67s memory: 249224k
        Processing time (w/o IO): 18.261384s
        total: 19.02s memory: 248968k
```

aardappel:
Oh, but those have improved? 189 -> 121ms for the first one?

Still surprising VS does better than Clang with this code, in my past benchmarking Clang was always ahead.

PR #430, opened 2023-11-14
[Typed Racket] single-threaded implementation

michaelsbradleyjr:
I adapted my work in `racket/related.rkt` to a single-threaded Typed Racket implementation.

Typed Racket is pretty sophisticated (my impression so far): great care is taken for code generated at untyped-typed boundaries so that at runtime there won't be a mismatch between what Typed Racket is working with (can involve some optimizations to memory layout, etc.) and the data passing to/from untyped sections.

That has some implications, though, for performance. For example: [`typed/json`](https://github.com/racket/typed-racket/blob/master/typed-racket-more/typed/json.rkt) is provided by the `typed-racket-more` library and is essentially just a wrapper around the untyped `json` library that's included in `racket/base`. So when naively consuming, say, a `(Listof JSExpr)` in Typed Racket code, runtime contracts (a pervasive feature of Racket) are involved in ensuring the data in the list is actually `JSExpr`. That it does that automatically is *really cool*, actually! But there's significant overhead and that's no good for this benchmark project.

So instead of using `typed/json` to read `posts.json`, I made use of the [`json-type-provider`](https://github.com/philnguyen/json-type-provider) library, which takes a different approach, i.e. without the overhead. However, it doesn't provider a `Writer` impl, only a `Reader`. It would be a great learning experience to implement `Writer` myself, but that's more than I want to take on at present; so, for now I've used `write-json` from `typed/json`. The step of serializing and writing data to `related_posts_typed_racket.json` is, therefore, quite pokey.

The runtime numbers for the step we care about (`(define related (process posts))`) are consistent with the untyped implementation. I had wondered if Typed Racket might do some additional optimizations resulting in a perf boost, but given the nature of the code (using `vector` and mutation) there's probably not much to optimize, but at least it's not slower than the untyped impl. 😄 

Because of the slow writes, I don't think it makes sense to include `run_typed_racket` in `all`. But I'll leave that for you to decide.

One more thing to note:

The `typed-racket-more` library is a dependency of `json-type-provider` (the latter uses the former's `JSExpr` type) and it's also needed for `typed/json`. A downside is that `typed-racket-more` is a hefty dependency with a lot of stuff that isn't needed for `typed/related.rkt`, or `json-type-provider` for that matter. It doesn't bloat the Docker image, but it does take a few minutes for `raco` to download and install all the modules pulled in by `typed-racket-more`. I also needed to install `cairo fontconfig pango` with pacman because some of the unneeded modules dynamically link to various `.so` libs.

Alternatively, I could extract and modify `.rkt` modules from `typed-racket-more` and `json-type-provider` and vendor them in `racket/typed/vendor/`, but I didn't do that yet. I can/will if you feel it's a blocker for this PR.


jinyus:
I guess we shouldn't add typed racket deps to docker since it'll not be included in the table. It's ok to just have it in the repo.

michaelsbradleyjr:
Okay, I'll comment them out.

michaelsbradleyjr:
> Okay, I'll comment them out.

Done, and I also added `--no-docs` to the enabled `raco pkg install` command since module docs don't need to be built.

michaelsbradleyjr:
@jinyus I forgot to change this line:

https://github.com/jinyus/related_post_gen/blob/add00ebe62569e5562fb620a3081aa83217d54a8/racket/info.rkt#L3C1-L3C45

With `"json-type-provider"` in the deps list, the `raco pkg install` step in `./run.sh racket` will take an unnecessarily long time.

I'll fix it.

michaelsbradleyjr:
I think I'll need to split out the Typed Racket impl from the Racket one, because `raco` is very thorough in checking what modules in a collection `(require ...)`.

I can't do that right at this moment but will take care of it asap, should be later today.

PR #432, opened 2023-11-15
[Racket] install racket instead of racket-minimal in Docker

michaelsbradleyjr:
This is *a* solution to the problem I realized/described [earlier today](https://github.com/jinyus/related_post_gen/pull/430#issuecomment-1811611857).

Advantages
* Having `"json-type-provider"` as a dep in `info.rkt` isn't a problem — `raco pkg install` won't be slow nor will it exit with error
* Typed Racket impl is runnable in the container, though still not included in `all` because of the slow writes
* Don't have to change anything else
* It becomes very unlikely that adding module dep/s for the Racket or Typed Racket impls (if they're to be runnable in the container) will also involve installing "missing pieces" via `pacman` and/or `raco` during `docker build`.

Disadvantages
* The size of the container increases by several hundred MB (but not a lot in the grand scheme of things).

Feel free to close this PR if it's not the solution you want, and I'll do another PR splitting the `racket` dir into `racket` and `typed_racket`, no problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,11 +86,7 @@ RUN pacman -S --noconfirm --needed leiningen stack
 RUN pacman -S --noconfirm --needed ruby
 
 # install racket
-RUN pacman -S --noconfirm --needed racket-minimal
-RUN raco pkg install --auto --no-docs compiler-lib
-# additional deps needed for Typed Racket, disabled for now
-# RUN pacman -S --noconfirm --needed racket-minimal cairo fontconfig pango
-# RUN raco pkg install --auto --no-docs compiler-lib typed-racket-more
+RUN pacman -S --noconfirm --needed racket
 
 # install lobster
 RUN pacman -S --noconfirm --needed cmake
--- a/run.sh
+++ b/run.sh
@@ -789,12 +789,12 @@ run_racket() {
         if [ -z "$appendToFile" ]; then # only build on 5k run
             raco pkg install --auto --name related --no-docs --skip-installed &&
                 raco make related.rkt
+        fi &&
+        if [ $HYPER == 1 ]; then
+            capture "Racket" hyperfine -r $runs -w $warmup --show-output "racket related.rkt"
+        else
+            command ${time} -f '%es %Mk' racket related.rkt
         fi
-    if [ $HYPER == 1 ]; then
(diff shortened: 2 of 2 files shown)


michaelsbradleyjr:
@jinyus I suspect 260033d was necessary because of the problem this PR is meant to address, which could involve the `raco pkg install` step exiting with error because of "missing pieces" (cairo `.so`, et al.), causing it to look like hyperfine wasn't used, but in fact it was simply not invoked. The 20k and 60k runs then proceed to work correctly because they don't invoke `raco pkg install`.

Note that `run_racket` in `run.sh` already invoked `hyperfine` prior to 260033d.

Sorry for any frustration and that I didn't realize the problem before you merged #430.

michaelsbradleyjr:
You should be able to revert 260033d and merge this PR, and it will work correctly, from what I observed in my Azure VM/Docker.

jinyus:
Yea, I only realized after a 1.5hr run. Will take a look at this PR after the next result update

michaelsbradleyjr:
> Yea, I only realized after a 1.5hr run
Ah, dang, really sorry about that 🤦 .

jinyus:
It's weird that it seems to be compiling unrelated stuff on the first run `webser,gui libs etc`. Would be great if that could be done in `docker build`

michaelsbradleyjr:
For sure, and that was part of #430, but then
you wrote:
> I guess we shouldn't add typed racket deps to docker since it'll not be included in the table. It's ok to just have it in the repo.

So I backed the Docker stuff out for #430 but forgot to account for the `typed-racket-more` dependency of `json-type-provider` in `racket/info.rkt`.

So then in one of your subsequent builds you hit the situation where `raco pkg install` is pulling in `typed-racket-more`, which has a crap-ton of dependencies; things went haywire and `raco pkg install` exited with error and it looked like `hyperfine` was missing.......

I've updated this PR to revert 260033d. As long as you're okay with `pacman` installing `racket` instead of `racket-minimal`, this PR is now good-to-go. If you're not okay with it, then I'll fix it a different way and promise not to screw up your builds doing so.

Again, apologies for the frustration/confusion. There were/are several variables in play and I'm pretty green with `raco` / Typed / Racket.

Issue #410, opened 2023-11-04
Raw pointers in C++ implementation


zigzag312:
Rules says no unsafe code blocks. Raw pointers are what other languages consider unsafe. Since smart pointers are available in modern C++, wouldn't replacing raw pointers with smart pointers make the most sense?

jinyus:
I'm not opposed to that but I'm no C++ expert so that will have to be contributed.

zigzag312:
My knowledge of C++ is rather light as well. I would change `RelatedPosts` struct like this:

```
struct RelatedPosts
{
    std::shared_ptr<std::string const> _id;
    std::shared_ptr<std::vector<std::string> const> tags;
    std::array<std::shared_ptr<Post const>, TOPN> related;
};
```
But then the use of `shared_ptr` needs to propagate all the way back to the input posts vector. Probably best to wait for someone with more C++ experience to chime in, so we don't create a Frankenstein's monster of an implementation.

cyrusmsk:
Btw how differ pointers in Nim, Go and Rust from C++ in that struct?

Go
```go
type RelatedPosts struct {
	ID      string      `json:"_id"`
	Tags    *[]string   `json:"tags"`
	Related [topN]*Post `json:"related"`
} 
```

Nim
```nim
RelatedPosts = ref object
    `"_id"`: string
    tags : ref seq[string]
    related: array[N, ptr Post]
```

Rust
```rust
struct RelatedPosts<'a> {
    #[serde(rename = "_id")]
    id: &'a str,
    tags: &'a [&'a str],
    related: Vec<&'a Post<'a>>,
}
```

all three are using pointers for tags array and post..

zigzag312:
I think Go's default pointers are managed by GC and are therefore safe.

Rust pointers have lifetime `'a` attached, so they are also safe?

I don't know about Nim.

jinyus:
According to https://nim-lang.github.io/Nim/tut1.html#advanced-types-reference-and-pointer-types, `ref` are GC managed and `ptr` is not, thus unsafe. That will have to be addressed. @michaelsbradleyjr

Issue #350, opened 2023-10-20
Allow .ptr if proven safe?


Imperatorn:
```d
		foreach (tag; post.tags)
			foreach (idx; tagMap[tag])
				taggedPostsCount[idx]++;
```
Should be able to be replaced by

```d
		foreach (tag; post.tags)
			foreach (idx; tagMap[tag])
				taggedPostsCount.ptr[idx]++;
```

Because it's proven safe as stated by the code.

"The size of taggedPostsCount is equal to the number of posts (postsCount). Since idx is generated from indices within the posts array itself (through the tagMap), idx will always be a valid index within taggedPostsCount. As such, you won't index taggedPostsCount with an out-of-bounds value."

Imperatorn:
Note that I am explicitly talking about out of bounds, not overflow.

taggedPostsCount can never be out of bounds by idx.

Hence I would argue that .ptr should be allowed in this case.

michaelsbradleyjr:
Fwiw, It seems reasonable to me.

I applied similar reasoning for what I did in `nim_con` with `addr`/`ptr`/`[]`. I haven't looked at your overall code, but in `nim_con` there is additional reasoning about safety because the input array (`seq` in Nim) and tag map are immutable after they are generated, and it can be clearly reasoned from the implementation that no two worker threads will ever attempt to write to the same index in the `seq` that, after the workers are finished, is serialized to JSON on the main thread and written to disk.

schveiguy:
FWIW, this is exactly the kind of real-world optimization that one would apply -- it's straightforward to reason about and explain in comments. I would do it in my code, because I know what care I would take to ensure it stays that way.

However, this cannot be proven at compile-time, and is very fragile. In the future, someone could mess up the storage of the indexes, and then you will have data corruption if the indexes go out of bounds. Or they might accidentally be used on the wrong array (in the concurrent version, we have a vector array which has a much smaller length).

The point of memory safety features such as bounds checks is to avoid having to check code for memory safety, even if you introduce bugs later. The rules in D are there for a reason, which means the bounds checks should stay in, unless the compiler can sleuth out that they are unnecessary (this does happen for the static arrays).

If I understand the goal of the safety requirement, I unfortunately think that .ptr cannot be acceptable in this instance. There is no specific safe interface to the code, it's just because we proved it *for that iteration of the code*.

If, however, there were a way to tie the indexing to a type, which can be proven to have a safe interface, then it should be allowed.

[...]

PR #468, opened 2023-11-21
[Nim] `array[N, ptr Post]` -> `array[N, Post]`

michaelsbradleyjr:
This PR removes use of `ptr` in the `nim` implementation, cf. https://github.com/jinyus/related_post_gen/issues/410#issuecomment-1819047087.

It is expected this change will make `nim` a little slower, but the point is to remove use of a language feature that Nim's [Manual](https://nim-lang.org/docs/manual.html) categorizes as unsafe.

From *[Nim Manual: The addr operator](https://nim-lang.org/docs/manual.html#statements-and-expressions-the-addr-operator)*
> An address is always an untraced reference. Taking the address of an object that resides on the stack is **unsafe**, as the pointer may live longer than the object on the stack and can thus reference a non-existing object.

The impact on performance seems to be minimal in AWS/Docker (I have a `c7a.xlarge` VM now). It's more noticeable in Azure/Docker, but the official benchmarking VM has been changed and per above it's expected anyway.

---

Does the use of `ptr` in `nim_con` need to be discussed. What about the use of `cast` in both `nim` and `nim_con`?

Maybe it wouldn't be a discussion *per se*, but @jinyus can make a decision/s. Once it's decided, I'll make whatever changes are needed for `nim` and/or `nim_con`, certainly without rancor. 😀 I understand the nature and intent of the rules re: *"unsafe"* and [this comment](https://github.com/jinyus/related_post_gen/issues/350#issuecomment-1773574145) has some good insights that apply to langs other than D, i.e. *"In the future, someone could mess up..."*

I have [already written](https://github.com/jinyus/related_post_gen/pull/291#issuecomment-1765170453) about this matter and related in several PR/issue comments. I will not repeat all of it, but instead attempt to boil it down.

From the *About* in Nim's home on GitHub 
> Nim's design focuses on efficiency, expressiveness, and elegance (in that order of priority).

Memory safety is not an afterthought in Nim's design, but it's not the primary focus. I don't think that will change much in another 10 or even 20 years. Rather, I think the Nim community's advice will be the same in the future as it is now (my personal impression): *Try to limit usage (quantity) of unsafe operations, and try restrict their usage so that humans can easily reason about those code sections.*

In other words, it's common that Nim programmers have to take memory safety into their own hands, albeit the importance of limiting the scope of doing so.

Pertinent to `nim_con`:

If you're doing an op where you need the compiler to e.g. interpret the bits of a `uint64` as an `int64`, then you need to use [`cast`, which the Manual categorizes as unsafe](https://nim-lang.org/docs/manual.html#statements-and-expressions-type-casts). That can be a reasonable thing to do if it's not math related, but a human has to make that judgement call, the compiler can't help you.

If you want to read/write data across threads — definitely possible in Nim v2 with ARC/ORC — then you need to use `ptr` (and locks as necessary). [Araq](https://github.com/Araq), the originator of Nim, even promotes `ptr` usage in his Nim v2 library [`malebolgia`](https://github.com/Araq/malebolgia).

The alternative to `ptr` for `nim_con` would be to take a no-shared approach and copy inputs/outputs across threads. It works, but last time I tried it the processing times were slower than `Java (GraalVM) Concurrent`, which has the slowest times in this project's *Multicore Results* scoreboard. No one evaluating/using Nim would take that hit in a *"real world"* project — they would either switch languages or make use of `ptr` with careful auditing.

For the sake of argument, let's assume `ptr` in `nim_con` and `cast` in `nim` and `nim_con` remain allowed because there aren't viable alternatives and (theoretically) we all agree such usage is effectively safe re: how it's done. But then reconsider usage of `ptr` in `nim` prior to this PR — it's even *more easily* reasoned that it's effectively safe because of how Nim's ARC/ORC memory model is documented to work.

I could elaborate, but I'm sure you get the idea at this point. I'll summarize this way: 

*With Nim you can write succint and effectively safe programs with top-tier performance, but it's not unlikely you will need to reason about memory safety yourself in some sections of your code, especially if your program is multi-threaded. If that's not acceptable to you or for your project's requirement, Nim may not be a good choice.*


jinyus:
I will allow `ptr` in nim_con and `cast` is fine in this instance.

PR #472, opened 2023-11-23
[rust] Various minor optimizations and some cleanups

cloneable:
* Look up tags by index.
* Reduce allocations and drop mimalloc.
* Use outside-loop vec (with good alignment) instead of arena allocator.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3,162 +3,25 @@
 version = 3
 
 [[package]]
-name = "arrayvec"
-version = "0.7.4"
+name = "equivalent"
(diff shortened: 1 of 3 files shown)


jinyus:
Hey, the rules mandate that tags must be string.
https://github.com/jinyus/related_post_gen/blob/93e2b14ee6fd63fa07354ae3ef1802b1a31ca838/readme.md?plain=1#L57

cloneable:
They are. They're read and written back out as strings.

jinyus:
I meant as map keys, I didn't want to get rid of hashing completely.

PR #477, opened 2023-11-26
[rust] various optimizations

cloneable:
* Drop bumpalo and mimalloc in favor an aligned vec.
* Use hashbrown instead of std+fxhasher.

PR #479, opened 2023-11-26
Revert "[rust] various optimizations"

jinyus:
Reverts jinyus/related_post_gen#477

https://github.com/rust-lang/hashbrown
```
SIMD lookups to scan multiple hash entries in parallel.
```
#### This is allowed for the concurrent implementation.

Rule violation:
https://github.com/jinyus/related_post_gen/blob/2a4eb605964a35ceaa837186867a6d9c2f8c73d0/readme.md?plain=1#L46

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3,28 +3,49 @@
 version = 3
 
 [[package]]
-name = "ahash"
-version = "0.8.6"
+name = "arrayvec"
+version = "0.7.4"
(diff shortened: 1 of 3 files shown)


jinyus:
The hashmap is the issue not aligned-vec

jinyus:
I think we should differentiate between facilitating SIMD and explicitly leveraging SIMD. Data alignment doesn't imply vectorization which is why I greenlit aligned-vec.

PR #505, opened 2024-01-19
Update Dockerfile

cyrusmsk:
LDC now updated in ArchLinux repo. No need to manually install 1.36 beta

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ FROM archlinux:base
 # Update package repository
 RUN pacman -Syu --noconfirm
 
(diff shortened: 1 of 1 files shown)


jinyus:
I actually prefer the explicitness of manual installation because we can rollback easily and we don't have to wait for the arch guys for updates.  I also plan to move all runtimes to manual installation when I have time.

cyrusmsk:
I see. Then maybe move from -beta to regular version. As 1.36 already published - https://github.com/ldc-developers/ldc/releases/tag/v1.36.0

PR #528, opened 2024-11-15
common lisp: update code, speedup

leetwinski:
it seems to work faster on my machine (~30-40% faster)

### DIFF
--- a/common-lisp/related.lisp
+++ b/common-lisp/related.lisp
@@ -1,112 +1,81 @@
-(declaim (optimize (speed 3) (safety 1)))
-
 ;; Install/load json lib
 (ql:quickload :com.inuoe.jzon)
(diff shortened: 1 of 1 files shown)


jinyus:
This is a little faster on the initial input but gets significantly slower as the input size grows.

Main:
```bash
| Common Lisp (SBCL) | 154.00 ms | 1.34 s | 11.20 s | 12.69 s |
```

This PR:
```bash
| Common Lisp (SBCL) | 142.00 ms | 4.03 s | 29.64 s | 33.81 s |
```

leetwinski:
Hm. that is weird. I'll take a look. Thank you!

zupat:
Closing this since #553 provides a great improvement

PR #529, opened 2025-02-10
optimize Haskell implementation by adopting GHC LLVM backend and full strictness

thelissimus:
- Switch GHC version from `9.8.1` to `9.10.1`
- Switch to LLVM backend
- Enable `Strict` extension
- `MutableByteArray` instead of `STRefU`
- `ShortText` instead of `Text` (1)
- `Storable` vector instead of `Primitive`
- Rewrite helpers as (tail-recursive) loops (2)
- Remove unnecessary warnings
- Remove bit packing
- Remove redundant standalone kind signatures

> 1. `ShortText` is actually more appropriate for the challenge since UTF-8 is enough (as opposed to `Text` with its UTF-16).
> 2. In `buildRelated` function one might see the usage of two **potentially** rule-breaking "unsafe" functions - `unsafeNew` and `unsafeFreeze`. But those names are deceptive and they are actually safe and therefore adhere to the rules. `unsafeNew` is just like `new` but doesn't initialize memory. Since we rewrite the contents anyways there is no need to initialize the vector when created. `unsafeFreeze` works in-place instead of copying the content like `freeze`. Since we won't use the accumulated mutable vector at all it is completely safe. No rules are broken.

Also, the build of `Dockerfile` might fail if there is no LLVM because Haskell implementation now depends on it. GHC supports versions from 13 up to 20 (not inclusive). [^1] Locally I used LLVM 15 and that worked fine. I wanted to test it locally but it has been building the image for hours now. The `Dockerfile` appears to include LLVM, but I haven’t verified compatibility. So, @jinyus, please double-check the Docker build.

[^1]: https://downloads.haskell.org/ghc/latest/docs/users_guide/codegens.html#llvm-code-generator-fllvm


thelissimus:
Also, you might want to stick around after merging this PR because I have another version ready that utilizes `brokkr-hashtables` instead of `vector-hashtables` which has 5-10% speed gains. First let's see how `vector-hashtables` performs then `brokkr-hashtables` then we will pick the fastest one for the task. (I can't rely on my local results to make the judgement, so it is better to just run it like the rest of the languages in the benchmarks and then decide)

jinyus:
Nice! This is much faster

Main:
```bash
| Haskell | 43.20 ms | 600.00 ms | 5.25 s | 5.89 s |
```

This PR:
```bash
| Haskell | 26.65 ms | 347.84 ms | 2.81 s | 3.19 s |
```

PR #508, opened 2024-02-07
[Regression] update to go 1.22.0

jinyus:
Single-core: 18% regression
Multi-core: TBA
Main:
```
| Go | 16.98 ms | 239.58 ms | 2.11 s | 2.36 s |
```

This PR:
```
| Go | 19.65 ms | 288.36 ms | 2.48 s | 2.79 s |
```

PR #542, opened 2025-07-21
Update Go version in Dockerfile to 1.24.5

jinyus:


jinyus:
Major regression: https://github.com/jinyus/related_post_gen/pull/508#issuecomment-3096747955

